PR #7, opened 2020-06-12
Simplify the grammars

billdenney:
This is perhaps not a real PR in that it may be best not to merge it.  This breaks detection of FIXED theta values, and if taken to its logical next step, it would also break detection of FIXED omega values.  It also breaks several other tests.  (To help orient me to the code, I would appreciate modifications here that show how to fix the detection of FIXED parameters.)

I think that it does improve detection of comments and spaces (as shown by the debug printing of the parse tree).  And, my goal with this was to ask a question:  Is it feasible and reasonable to detect parameter names via comments as was done with PsN?  Perhaps as a first step in that direction, could the comments following parameter declaration be added to the object as a new field?

### DIFF
--- a/src/pharmpy/plugins/nonmem/records/grammars/definitions.lark
+++ b/src/pharmpy/plugins/nonmem/records/grammars/definitions.lark
@@ -1,8 +1,15 @@
-WS: (" " | /\t/)+
-WS_ALL: /\s+/
+// Blanks and ignored characters
 
-NEWLINE: (/\r?\n[ \t]*/ | COMMENT)+
+blank : ( WS | newline )
+
+newline : ( WS* COMMENT? NEWLINE )+
+
+WS_ALL: WS
+WS: /[ \t\f\v]+/
+NEWLINE: (/\r?\n/)+
 COMMENT: /;[^\n]*/
+
+// Numbers
 
 DIGIT: "0".."9"
 INT: DIGIT+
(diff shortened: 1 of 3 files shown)


rikardn:
Thanks for the PR! I think that it would be easiest, and most flexible, to leave the parameter name comment as a comment by the parser and then let code do the extraction of the parameter name. The plan is to add an option to either use the NONMEM names or to use the PsN-style comments. Also NONMEM 7.5 have a new way of naming parameters that would be interesting to support.

As for your cleanup, I might not have time to review until after my vacation

billdenney:
Leaving the comments as comments makes sense to me.  I'm also looking forward to the NONMEM 7.5 parameter naming (if we're talking about the same thing, I'm the person who suggested it to Bob :) ).

One complexity of associating comments with parameters that this PR would not work for is if you have comments on the next line after the parameter, too, I think that this PR would have multiple `COMMENT` terminals.  Mainly, these are the scenarios that would need to be distinguished:

# Comments on the same line as the object
```
$THETA
(1, 2, 3) ; I am a comment
```

# Comments before, on the same line, and after the object
```
$THETA
; I am a comment before
(1, 2, 3) ; I am a comment on the same line
; I am a comment on the next line
```

# Are comments within the object legal (I would guess that NMTRAN parses them as legal, but I have not confirmed that)
```
$THETA
; I am a comment before
(1, ; I am a comment within the object, why would anyone do this?!  (but someone probably has...)
2, 3) ; I am a comment on the same line
; I am a comment on the next line
```

rikardn:
I have now implemented using labels as parameter names. The implementation ended up using the current grammar and programmatically search for the next comment after each initial estimate. Default is still to use the "standard" NONMEM names, but there is an option to change this. I haven't written documentation for this yet, but under Linux the config file is .config/Pharmpy/pharmpy.conf and here is the content to change the default behaviour:

```
[pharmpy.plugins.nonmem]
parameter_names=comment
```

Issue #67, opened 2020-12-18
Parser Fails with Symbolic Thetas


billdenney:
I have a model that uses symbolic `THETA()` values with `$ABBR REPLACE`.  I was playing around with pharmpy via R, and I got an error where it appears that the parser was expecting an integer.

I had to subtly change the message so that it masks the underlying model:

```r
> models[[1]]$to_base_model()
 Error in py_call_impl(callable, dots$args, dots$keywords) : 
  UnexpectedCharacters: No terminal defined for 'L' at line 7 col 23

  LTVCL   = THETA(LTVCL)
                  ^

Expecting: {'WS', 'INTEGER'} 
```

The model has a line like the following

```
$ABBR REPLACE THETA(LTVCL)=THETA(1)
```

rikardn:
Thanks for reporting this. Pharmpy doesn't yet support `$ABBR`. I don't think it would be too difficult to add support though.

rikardn:
A first step 2b3333090acc516f1bfe5ee9163541cea6a4eac5 handles parsing of `$ABBR`.

stellabelin:
Configuration is now possible to parse and write $ABBR (5b171074403170eef84aafce81f6a06734c19ce1 and feceb126d0aa5b7e376d8d831d97439f01652360)

PR #68, opened 2020-12-18
Example abbreviated THETA and ETA models

billdenney:
This is related to #67.

Here are two variants of the pheno model that cause the parser to fail but work when run with `nmfe75`.  The R code below illustrates the issue.  (Modify `path_to_abbr` to match where the files are on your system.)

```r
library(reticulate)
pharmpy <- import("pharmpy")

path_to_abbr <- "c:/tmp"
model_files_pk <- list.files(path=path_to_abbr, pattern="*mod$", full.names=TRUE)
models <-
  lapply(
    X=setNames(model_files_pk, basename(model_files_pk)),
    FUN=pharmpy$Model
  )
models[[1]]$statements
models[[2]]$statements
models[[3]]$statements
```


rikardn (approved):
Thanks @billdenney. This is very helpful!

PR #402, opened 2021-11-18
Change the best model selection criteria

ZheHuang96:
Hi, the request is about changing the best model selection criteria in resmod. Because in this case, the "dofv" is always negative and we need to select the lowest one.


rikardn (approved):

rikardn:
Thanks!

PR #408, opened 2021-11-21
Final estimates in resmod

ZheHuang96:
Hi, here are two commits for final estimates in resmod. The first one is to add the final estimates into the resmod models data frame as parameters, the second one is to use the parameters as the initial estimates for setting up full models.

### DIFF
--- a/src/pharmpy/tools/resmod/results.py
+++ b/src/pharmpy/tools/resmod/results.py
@@ -1,6 +1,5 @@
 from pathlib import Path
 
-import numpy as np
 import pandas as pd
 
 from pharmpy.results import Results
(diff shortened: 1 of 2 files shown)


rikardn:
Thanks!

The linter check doesn't pass. Could you run `tox -e check` and then fix the raised issues. Then you can amend the changes to the same commit with `git commit --amend` and force push `git push -f origin final-estimates`

src/pharmpy/tools/resmod/results.py
```diff
@@ -17,13 +16,15 @@ def calculate_results(base_model, iiv_on_ruv, power):
     base_ofv = base_model.modelfit_results.ofv
     dofv_iiv_on_ruv = iiv_on_ruv.modelfit_results.ofv - base_ofv
     dofv_power = power.modelfit_results.ofv - base_ofv
+    omega_iiv_on_ruv = iiv_on_ruv.modelfit_results.parameter_estimates[-2]
```
rikardn, line 19: Even though it would work in this case I wouldn't use numerical indices. Better is to index directly using the name of the parameter

src/pharmpy/tools/resmod/tool.py
```diff
@@ -133,12 +133,14 @@ def _create_dataset(input_model):
...
 def _create_best_model(model, res):
     model = model.copy()
-    if any(res.models['dofv'] > 3.84):
-        idx = res.models['dofv'].idxmax()
+    if any(-res.models['dofv'] > 3.84):
+        idx = res.models['dofv'].idxmin()
         name = idx[0]
         if name == 'power':
             set_power_on_ruv(model)
+            model.parameters.inits = {'power1': res.models['parameters'].iloc[1].get('theta')}
```
rikardn, line 141: `iloc[1]` is dangerous since it might change in the future. Instead use the row-index of the `res.models`.  For example: `res.models['parameters'].loc['power', 1, 1]`

rikardn (changes requested):

src/pharmpy/tools/resmod/results.py
rikardn: Thanks for fixing this.

rikardn:
Thanks this is looking good to me. I will merge when the test have passed.

rikardn (approved):

PR #436, opened 2021-12-01
Add method insert_before to ModelStatements

rikardn:


src/pharmpy/statements.py
```diff
@@ -1776,6 +1776,29 @@ def full_expression(self, expression):
             expression = expression.subs({statement.symbol: statement.expression})
         return expression
 
+    def insert_before(self, statement, new_statement):
+        """Insert a statement just before another statement
+
+        Parameters
+        ----------
+        statement : Statement
```
stellabelin, line 1784: Since there is no Statement class, this should be `statement : Assignment or ODESystem`

src/pharmpy/statements.py
```diff
@@ -1776,6 +1776,29 @@ def full_expression(self, expression):
...
+            Insert before this statement
+        new_statement : Statement
```
stellabelin, line 1786: Same as previous

src/pharmpy/statements.py
rikardn: Ok, I'll add a Statement base class. This could also be used for `isinstance` checks.

src/pharmpy/statements.py
rikardn: Will be resolved with the above

rikardn:
I have added an abstract base class for all Statements. This is ready for review again.

stellabelin (approved):

PR #440, opened 2021-12-05
Resmod

ZheHuang96:
Hi Rikard,

Here are four commits related to Resmod tool. Because I can not rebase my fork successfully, I copied your commit 'Let resmod use CIPREDI if available' for tool/resmod/tool.py and create the dataset based on your change. 

Thank you!

Best regards,
Zhe Huang


ZheHuang96:
I am so sorry that there is a conflict that need to be resolved and it is the create dataset part in resmod/tool.py. Could you please tell me what should I do in this case? Tack!

src/pharmpy/modeling/data.py
```diff
@@ -205,3 +206,29 @@ def get_observations(model):
...
+    -------
+    pd.Series
+        MDVs
+    """
+    try:
+        label = model.dataset.pharmpy.labels_by_type[ColumnType.EVENT]
+    except KeyError:
+        try:
+            label = model.dataset.pharmpy.labels_by_type[ColumnType.DOSE]
+        except KeyError:
```
rikardn, line 229: If no event or dose column. Set all rows to MDV=0 (all rows are observations by default)

rikardn (approved):

PR #444, opened 2021-12-08
Add zero values for mdv when only observations

ZheHuang96:
Hi Rikard,

I add the series of zero mdv when there is no dose in the dataset. I use the length of DV to create the series, I am not sure it is the best way or not...

### DIFF
--- a/src/pharmpy/modeling/data.py
+++ b/src/pharmpy/modeling/data.py
@@ -1,4 +1,8 @@
 # Functional interface to extract dataset information
+import numpy as np
+import pandas as pd
+
+from pharmpy.data import ColumnType
 
 
 def get_number_of_individuals(model):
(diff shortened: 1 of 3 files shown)


src/pharmpy/modeling/data.py
```diff
@@ -227,8 +231,12 @@ def get_mdv(model):
...
             label = model.dataset.pharmpy.labels_by_type[ColumnType.DOSE]
         except KeyError:
-            raise DatasetError('Could not find EVID or AMT column in dataset')
+            label = model.dataset.pharmpy.labels_by_type[ColumnType.DV]
+
+    data = model.dataset[label].astype('float64').squeeze()
 
-    dose = model.dataset[label].astype('float64').squeeze()
-    mdv = dose.where(dose == 0, other=1).astype('int64').rename('MDV')
+    if label[0] == 'DV':
```
rikardn, line 238: You shouldn't assume that the column name is 'DV' here. Handle this case directly in the except and then return.

rikardn (changes requested):

src/pharmpy/modeling/data.py
rikardn: Now it looks better. Thanks!

rikardn (approved):

PR #465, opened 2021-12-16
Add id column in resmod dataset

ZheHuang96:
Hi Rikard,

This commit is to fix the issue that we have different OFV values in Resmod of Pharmpy and PsN.


rikardn (approved):
LGTM!

PR #469, opened 2021-12-20
Add combined error model in resmod

ZheHuang96:
Hi Rikard,

Here is the combined error model. This commit included combined error model in resmod and the original model transformation when combined error model is the best resmod model. I will force IPRED equal to a small value when it is zero in the next commit:)

### DIFF
--- a/src/pharmpy/tools/resmod/tool.py
+++ b/src/pharmpy/tools/resmod/tool.py
@@ -1,15 +1,17 @@
 import pandas as pd
+import sympy
 
 import pharmpy.model
 import pharmpy.tools
 from pharmpy import Parameter, Parameters, RandomVariable, RandomVariables
 from pharmpy.data import ColumnType
 from pharmpy.estimation import EstimationStep, EstimationSteps
-from pharmpy.modeling import get_mdv, set_iiv_on_ruv, set_power_on_ruv
+from pharmpy.modeling import get_mdv, set_combined_error_model, set_iiv_on_ruv, set_power_on_ruv
 from pharmpy.statements import Assignment, ModelStatements
 from pharmpy.tools.modelfit import create_fit_workflow
 from pharmpy.workflows import Task, Workflow
 
+from ...modeling.error import remove_error_model
 from .results import calculate_results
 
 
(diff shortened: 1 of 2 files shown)


src/pharmpy/tools/resmod/tool.py
```diff
@@ -121,6 +125,43 @@ def _create_power_model(input_model):
     return model
 
 
+def _create_combined_error_model(input_model):
+    combined_error_model = pharmpy.model.Model()
```
rikardn, line 129: Does this need to be created from scratch? Couldn't it use the base model as starting point?

ZheHuang96: Sorry, I never think about this point...I can have a try!

src/pharmpy/tools/resmod/tool.py
```diff
@@ -121,6 +126,33 @@ def _create_power_model(input_model):
     return model
 
 
+def _create_combined_model(input_model):
```
rikardn, line 129: This is complicating things unnecessarily. No need to be general here. There is only one assignment in the model. Add a new epsilon / IPRED to that expression.

rikardn: Something like:

```
s = model.statements[0]
s.expression = s.expression + new_ruv / IPRED
```
The init for the original sigma can be changed:

```
model.parameters['sigma'].init = new_value
```

The new epsilon and sigma needs to be added the way you do it.

src/pharmpy/tools/resmod/tool.py
```diff
@@ -121,6 +126,33 @@ def _create_power_model(input_model):
...
+    base_model = input_model
+    model = base_model.copy()
+    remove_error_model(model)
+    stats, y, f = _preparations(model)
```
rikardn, line 133: _preparations is a private function so it shouldn't be used here. Given my previous comment it shouldn't be needed.

src/pharmpy/tools/resmod/tool.py
ZheHuang96: Hi Rikard, no problem, I can change it. But I still have a question, do we need to specify the proportional sigma and additive sigma so that we can use the final estimates of these two sigmas to be initial values of the transformed best model. If we need, can I change the original sigma's name from 'sigma' to 'sigma_prop'? I don't know how to change the name and symbol of the original sigma so I delete it and add two new sigma/epsilon.

src/pharmpy/tools/resmod/tool.py
ZheHuang96: Yes, thank you! Now I understand more about the statements:)

src/pharmpy/tools/resmod/tool.py
rikardn: Good point. I think your new version is good.

rikardn (approved):

PR #509, opened 2022-01-18
Add time varying error model function

ZheHuang96:
Hi Rikard, 

Here is the add_time_varying_error function. After this task, I am going to change the models and workflow in resmod:)

### DIFF
--- a/src/pharmpy/modeling/error.py
+++ b/src/pharmpy/modeling/error.py
@@ -618,7 +618,7 @@ def set_time_varying_error_model(model, cutoff, idv='TIME'):
 
     Examples
     --------
-    >>> from pharmpy.modeling import load_example_model, add_time_varying_error_model
+    >>> from pharmpy.modeling import load_example_model, set_time_varying_error_model
     >>> model = load_example_model("pheno")
     >>> set_time_varying_error_model(model, cutoff=1.0)    # doctest: +ELLIPSIS
     <...>
--- a/src/pharmpy/modeling/error.py
+++ b/src/pharmpy/modeling/error.py
@@ -597,3 +597,49 @@ def set_dtbs_error_model(model, fix_to_log=False):
     model.observation_transformation = obs
 
     return model
+
+
+def set_time_varying_error_model(model, cutoff, idv='TIME'):
+    """Set a time varying error model per time cutoff
+
+    Parameters
+    ----------
+    model : Model
+        Pharmpy model
+    cutoff : float
+        A value at the given quantile over idv column
+    idv : str
+        Time or time after dose, default is Time
+
+    Return
+    ------
+    Model
+        Reference to the same model object
+
+    Examples
+    --------
+    >>> from pharmpy.modeling import load_example_model, add_time_varying_error_model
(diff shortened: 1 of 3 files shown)


src/pharmpy/modeling/error.py
```diff
@@ -597,3 +597,47 @@ def set_dtbs_error_model(model, fix_to_log=False):
...
+
+
+def add_time_varying_error_model(model, cutoff, idv='TIME'):
+    """Set a time varying error model per time cutoff
+
+    Parameters
+    ----------
+    model : Model
+        Pharmpy model
+    cutoff : Float
```
rikardn, line 609: `Float` should be `float`

src/pharmpy/modeling/error.py
```diff
@@ -597,3 +597,47 @@ def set_dtbs_error_model(model, fix_to_log=False):
...
+        A value at the given quantile over idv column
+    idv : String
```
rikardn, line 611: `String` should be `str`  (`str` being the name of the string class)

src/pharmpy/modeling/error.py
```diff
@@ -597,3 +597,47 @@ def set_dtbs_error_model(model, fix_to_log=False):
...
+        Time or time after dose, default is Time
+
+    Return
+    ------
+    Model
```
rikardn, line 616: Add the explanation line of the Model as in the other functions

src/pharmpy/modeling/error.py
```diff
@@ -597,3 +597,47 @@ def set_dtbs_error_model(model, fix_to_log=False):
     model.observation_transformation = obs
 
     return model
+
+
+def add_time_varying_error_model(model, cutoff, idv='TIME'):
```
rikardn, line 602: Perhaps we should rename it to `set_time_varying_error_model` as we actually change the model to become time varying.

rikardn (changes requested):
Looks good in general. Some minor fixes needed.

rikardn:
Also check the failing doctest. Seems to be a space too much in the output given in the docstring

ZheHuang96:
No problem, I can fix the problems now. I was checking the failing doctest for py39 and feel confusing about it... Thank you for your comment, I am going to check the space of the output!

ZheHuang96:
I am sorry I press the wrong button

src/pharmpy/modeling/error.py
```diff
@@ -597,3 +597,49 @@ def set_dtbs_error_model(model, fix_to_log=False):
...
+        Reference to the same model object
+
+    Examples
+    --------
+    >>> from pharmpy.modeling import load_example_model, set_time_varying_error_model
+    >>> model = load_example_model("pheno")
+    >>> set_time_varying_error_model(model, cutoff=1.0)    # doctest: +ELLIPSIS
+    <...>
+    >>> model.statements.find_assignment("Y")
+          ⎧EPS(1)⋅F⋅time_varying + F  for TIME < 1.0
```
rikardn, line 626: One space too much for the output

rikardn (approved):
Thanks! This is looking good now.

PR #510, opened 2022-01-20
Time varying error resmod

ZheHuang96:
Hi Rikard,

Here is time varying error in Resmod. I tried my best to make the code general but I still have some questions on this code. For example, now we give a default: groups = 4, so I know it will generate 3 models, so I add three models in `calculate_results` function of results.py and `_create_best_model` of tool.py directly. I guess maybe it is possible to let the models change according to the groups number we give automatically by loop but I haven't found a way to do that. Maybe you can have a look at my code and give me some comments first? So far it should work for adding 3 time varying error models in Resmod. 

Thank you very much!

Best regards,
Zhe Huang

### DIFF
--- a/src/pharmpy/tools/resmod/results.py
+++ b/src/pharmpy/tools/resmod/results.py
@@ -12,7 +12,7 @@ class ResmodResults(Results):
         self.models = models
 
 
-def calculate_results(base_model, iiv_on_ruv, power, combined):
+def calculate_results(base_model, iiv_on_ruv, power, combined, tvar_models):
     base_ofv = base_model.modelfit_results.ofv
     dofv_iiv_on_ruv = base_ofv - iiv_on_ruv.modelfit_results.ofv
     dofv_power = base_ofv - power.modelfit_results.ofv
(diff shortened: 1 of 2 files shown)


src/pharmpy/tools/resmod/results.py
```diff
@@ -12,7 +12,9 @@ def __init__(self, models=None):
         self.models = models
 
 
-def calculate_results(base_model, iiv_on_ruv, power, combined):
+def calculate_results(
```
rikardn, line 15: Use a list of time varying models as input

src/pharmpy/tools/resmod/results.py
```diff
@@ -21,16 +23,38 @@ def calculate_results(base_model, iiv_on_ruv, power, combined):
     theta_power = round(power.modelfit_results.parameter_estimates["power1"], 6)
     sigma_add = round(combined.modelfit_results.parameter_estimates["sigma_add"], 6)
     sigma_prop = round(combined.modelfit_results.parameter_estimates["sigma_prop"], 6)
-
+    dofv_tvar1 = base_ofv - time_varying1.modelfit_results.ofv
```
rikardn, line 26: Loop over all tvar models

src/pharmpy/tools/resmod/tool.py
```diff
@@ -39,8 +42,18 @@ def create_workflow(model=None):
     task_combined = Task('create_combined_error_model', _create_combined_model)
     wf.add_task(task_combined, predecessors=task_base_model)
 
-    fit_wf = create_fit_workflow(n=4)
-    wf.insert_workflow(fit_wf, predecessors=[task_base_model, task_iiv, task_power, task_combined])
+    tasks = []
+    groups = 4
```
rikardn, line 46: Have groups as an argument to create_workflow with default 4

src/pharmpy/tools/resmod/tool.py
```diff
@@ -67,6 +80,9 @@ def post_process(start_model, *models):
         iiv_on_ruv=_find_model(models, 'iiv_on_ruv'),
         power=_find_model(models, 'power'),
         combined=_find_model(models, 'combined'),
+        time_varying1=_find_model(models, 'time_varying1'),
```
rikardn, line 83: Use a loop or create new function to find all time varying models into a list

src/pharmpy/tools/resmod/tool.py
```diff
@@ -194,6 +240,27 @@ def _create_best_model(model, res):
             model.parameters.inits = {
                 'IIV_RUV1': res.models['parameters'].loc['IIV_on_RUV', 1, 1].get('omega')
             }
+        elif name == 'time_varying1':
```
rikardn, line 243: Loop over all previously found tvar models

rikardn (changes requested):
The main issue is that you should use loops and lists for the cases where there is an unknown number of models

ZheHuang96:
Hi Rikard,

Thank you so much for your comments! I have rewritten the code with loops. But for the create the best model part, I guess maybe we don't need a loop because if dofv>3.84, we only can have one model with max dofv. If it is a time varying model, my idea is to use this model's name to implement. Because I give a different name to different time varying model per cutoff, e.g. 'time_varying1' for the first cutoff, etc. 

Best regards,
Zhe Huang

rikardn (approved):
Now I think it looks good

ZheHuang96:
Thank you:)

PR #514, opened 2022-01-24
Zero protection

ZheHuang96:
Hi Rikard,

Here is the zero protection for power model and special combined model in resmod. I think for combined error, maybe the easier way is to avoid zero IPRED in dataset. I also tried to change the expression of combined model but it seems more complicated and I haven't figured it out. If it is the better way, I can continue working on it :) 

Best regards,
Zhe Huang


rikardn:
LGTM, but could you check out the failing doctest

ZheHuang96:
Yes, I am checking now.

PR #518, opened 2022-01-26
Change best model selection criteria with p=0.01 in resmod

ZheHuang96:
Hi Rikard,

Mats suggested using p=0.01 instead of p=0.05 as the selection criteria for resmod models, so I changed the cutoff in resmod.

Best regards,
Zhe Huang


rikardn:
Thanks!

PR #525, opened 2022-02-06
Resmod

ZheHuang96:
Hi Rikard,

Here are two commits for resmod. Mats and Xiaomei suggest comparing different significant levels in resmod, so I add an option to change the cutoffs by users. The other commit is to avoid unnecessary tasks when running resmod multiple times and make the resmod model results data frame flexible.

Best regards,
Zhe Huang

### DIFF
--- a/src/pharmpy/tools/resmod/tool.py
+++ b/src/pharmpy/tools/resmod/tool.py
@@ -23,7 +23,7 @@ from ...modeling.error import remove_error_model, set_time_varying_error_model
 from .results import calculate_results
 
 
-def create_workflow(model=None, groups=4, cutoff=3.84):
+def create_workflow(model=None, groups=4, cutoff=3.84, skip=[]):
     wf = Workflow()
     wf.name = "resmod"  # FIXME: Could have as input to Workflow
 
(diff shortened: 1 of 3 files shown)


src/pharmpy/tools/resmod/tool.py
```diff
@@ -35,26 +35,31 @@ def create_workflow(model=None, groups=4):
...
-        fit_wf, predecessors=[task_base_model, task_iiv, task_power, task_combined] + tasks
-    )
-
-    task_post_process = Task('post_process', post_process)
+    if 'IIV_on_RUV' not in skip:
+        task_iiv = Task('create_iiv_on_ruv_model', _create_iiv_on_ruv_model)
+        tasks.append(task_iiv)
+        wf.add_task(task_iiv, predecessors=task_base_model)
+
+    if 'power' and 'combined' not in skip:
```
rikardn, line 44: This is not what you intend. It will be testing `'combined' not in skip` only.

rikardn (changes requested):
Just one error.

src/pharmpy/tools/resmod/tool.py
ZheHuang96: Thank you for your comment! I forgot it...I correct the code now!

ZheHuang96: It should be `if 'power' not in skip and 'combined' not in skip:`

rikardn:
LGTM

PR #530, opened 2022-02-11
Set combined error expression when input model has extended errors

ZheHuang96:
Hi Rikard,

This commit is to change the combined error model expression for models that have three extended errors. The three extended error are: iiv_on_ruv, time_varying and iiv_on_ruv+time_varying.
I found a way to get specific expression and condition from a piecewise function in Stackoverflow and it works! The codes I write in line 331-333 look very weird but I had to substitute the previous proportional epsilon in the input model with the new proportional epsilon, otherwise, I will have three epsilons in the random variables(previous proportional epsilon, new proportional epsilon, new additive epsilon ), then the nonmem code cannot generate the correct expression and have `EPS(3)` in the code. I guess maybe there is a better way to do it? 

Thank you!

Best regards,
Zhe Huang

### DIFF
--- a/src/pharmpy/modeling/error.py
+++ b/src/pharmpy/modeling/error.py
@@ -311,18 +311,52 @@ def set_combined_error_model(model, data_trans=None):
     if has_combined_error_model(model):
         return model
     stats, y, f = _preparations(model)
+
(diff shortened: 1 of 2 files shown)


ZheHuang96:
Thank you for your reply!

PR #556, opened 2022-02-18
Check if initial estimates are valid in Model parameter setter

stellabelin:


src/pharmpy/model.py
```diff
@@ -109,9 +111,40 @@ def parameters(self):
...
+            nearest = self.random_variables.nearest_valid_parameters(inits)
+            if nearest != inits:
+                before, after = self._compare_before_after_params(inits, nearest)
+                warnings.warn(
+                    f"Adjusting initial estimates to create positive semidefinite "
+                    f"omega/sigma matrices.\nBefore adjusting:  {before}.\n"
+                    f"After adjusting: {after}"
+                )
+                self._parameters.inits = nearest
+                return
```
rikardn, line 130: If value is Parameters cannot return here. Only inits have been updated

rikardn (changes requested):

src/pharmpy/model.py
stellabelin: Have changed to: 
```
if isinstance(value, Parameters):
   value.inits = nearest
else:
   value = nearest
```

rikardn (approved):

PR #566, opened 2022-02-25
Remove iiv/iov

ZheHuang96:
Hej Rikard, 

Here are the commits related to` remove_iiv() ` and `remove_iov()`.  In the `remove_iov()` case, based on the IOV code that pharmpy generated, when we remove specific IOV etas, it should be a group of etas that have the same omega variance. So I think it is maybe better to use the omega name as the input option to remove all etas. And we could remove specific IOV assignment by inputting the IOV symbol name. 

Best regards,
Zhe Huang


src/pharmpy/modeling/remove_iov.py
```diff
@@ -4,16 +4,25 @@
...
 
-def remove_iov(model):
+def remove_iov(model, to_remove_parameters=None, to_remove_symbols=None):
     """Removes all IOV etas
 
     Parameters
     ----------
     model : Model
         Pharmpy model to remove IOV from.
+    to_remove_parameters : str, list
```
rikardn, line 20: It should be a list of individual parameters or random variables like the iiv case. If a random variable is selected perhaps all in the same iov group could be removed.

src/pharmpy/modeling/remove_iov.py
```diff
@@ -4,16 +4,25 @@
...
+        Name/names of omega parameters for IOV etas to remove, e.g. 'OMEGA_IOV_1'.
+        If None, all etas that are IOVs and related parameters will be removed. None is default.
+    to_remove_symbols : str, list
```
rikardn, line 23: The IOV assignments are really arbitrary. I think this option should be removed.

rikardn (changes requested):

src/pharmpy/modeling/remove_iov.py
ZheHuang96: > The IOV assignments are really arbitrary. I think this option should be removed.

Ok, I can delete this option!

src/pharmpy/modeling/remove_iov.py
rikardn: I mean: as if we imagine all to be the same ETA and can be represented with any one ETA in the same group.

ZheHuang96: > I mean: as if we imagine all to be the same ETA and can be represented with any one ETA in the same group.

Ok, let me use this code as an example:
```
$PK
IOV_1 = 0
IF (VISI.EQ.3) IOV_1 = ETA(4)
IF (VISI.EQ.8) IOV_1 = ETA(5)
IOV_2 = 0
IF (VISI.EQ.3) IOV_2 = ETA(6)
IF (VISI.EQ.8) IOV_2 = ETA(7)
IOV_3 = 0
IF (VISI.EQ.3) IOV_3 = ETA(8)
IF (VISI.EQ.8) IOV_3 = ETA(9)
ETAI1 = IOV_1 + ETA(1)
ETAI2 = IOV_2 + ETA(2)
ETAI3 = IOV_3 + ETA(3)
CL = THETA(1)*EXP(ETAI1)
VC = THETA(2)*EXP(ETAI2)
MAT = THETA(3)*EXP(ETAI3)
KA = 1/MAT
V = VC
$ERROR
CONC = A(2)/VC
Y = CONC + CONC * EPS(1)
$ESTIMATION METHOD=1
$THETA (0, 10.0, Inf) ; POP_CL
$THETA (0, 100, Inf) ; POP_VC
$THETA (0, 3.0, Inf) ; POP_MAT
$OMEGA 0.1; IIV_CL
$OMEGA 0.1; IIV_VC
$OMEGA 0.1; IIV_MAT
$OMEGA  BLOCK(1)
0.010000000000000002 ; OMEGA_IOV_1
$OMEGA  BLOCK(1) SAME ; OMEGA_IOV_1
$OMEGA  BLOCK(1)
0.010000000000000002 ; OMEGA_IOV_2
$OMEGA  BLOCK(1) SAME ; OMEGA_IOV_2
$OMEGA  BLOCK(1)
0.010000000000000002 ; OMEGA_IOV_3
$OMEGA  BLOCK(1) SAME ; OMEGA_IOV_3
$SIGMA 0.1; RUV_PROP
```
In this model we have 6 iov etas :
Skärmbild 2022-02-25 143540
So what we want is to input something like `ETA_IOV_11`, then remove both `ETA_IOV_11` and `ETA_IOV12`?

rikardn: Yes, that was the idea. `['ETA_IOV_11']`, `['ETA_IOV_12']`  and `['ETA_IOV11', 'ETA_IOV_12']`  should all remove both etas in the block.

ZheHuang96: > Yes, that was the idea. `['ETA_IOV_11']`, `['ETA_IOV_12']` and `['ETA_IOV11', 'ETA_IOV_12']` should all remove both etas in the block.

Ok, I see. My first idea is using the iov eta names as input to remove because the way we named iov is different from iiv, so I feel it is hard to use the individual parameters as input as when we did in `remove_iov()`. But then I found, for users, it may be also hard to get the right names of iov eta, because after `add_iov() ` function, we got the code as I posted above. The eta names are not shown in the code, instead, we had the omega names as comments, so I can use the omegas(eta_parameter_names) to find which group of etas we should remove. So that is what I implement here.

I can go back to the eta names idea and try to use them as indicators :)

rikardn: The eta names can be found with `model.random_variables`. IOV etas in the same group could be found assuming they have the same omega parameters (`free_symbols` would perhaps give this). The API of the modeling functions should target the Pharmpy representation of the model. The NONMEM implementation is something that comes in the update part. We also have a mechanism to save eta names in the NONMEM model via the `$ABBR` record. There is a confguration option to activate this.

ZheHuang96: > The eta names can be found with `model.random_variables`. IOV etas in the same group could be found assuming they have the same omega parameters (`free_symbols` would perhaps give this). The API of the modeling functions should target the Pharmpy representation of the model. The NONMEM implementation is something that comes in the update part. We also have a mechanism to save eta names in the NONMEM model via the `$ABBR` record. There is a confguration option to activate this.

Thank you so much for your detailed reply! I understand it better now. I will have a try:)

ZheHuang96:
Hi Rikard,

In this update PR, `remove_iov()` can remove all the iov etas in the same block by inputting any combinations of etas with the same omega variance. I see the check failed, I will update my python version and do PR again.

rikardn:
The tests failed because the CI was temporarily broken. I have fixed it now.

ZheHuang96:
Thank you for your information! I am not sure how to restart the check, so I add python 3.10 into the test environment of tox to see the commit can pass the checks or not. I can remove this commit later:)

rikardn:
Thanks for the updates!

PR #576, opened 2022-03-01
Add descriptor to ColumnInfo

rikardn:


src/pharmpy/datainfo.py
```diff
@@ -52,6 +54,7 @@ class ColumnInfo:
         'nmtran-time',
         'nmtran-date',
     ]
+    _all_descriptors = [None, 'body weight']
```
stellabelin, line 57: Should `age` also be in `_all_descriptors` since it is in the docstring?

rikardn: Yes, missed it. Good catch

PR #591, opened 2022-03-04
New iiv version of BIC in calculate_bic (fixes #580)

### LINKED ISSUE #580: Add option to calculate iiv-version of BIC

ZheHuang96:
Hi Rikard,

Here is the commit for the new BIC formula for IIV, I read the related paper again and I think my understanding of the formula is OK, but I can double-check with others on the Monday meeting so that I would not make the mistake.

Have a good weekend!

Best regards,
Zhe Huang


src/pharmpy/modeling/results.py
```diff
@@ -562,7 +562,8 @@ def calculate_bic(model, type=None):
       | BIC = OFV + n_estimated_parameters * log(n_observations)
     * | random
       | BIC = OFV + n_estimated_parameters * log(n_individals)
-
+    * | iiv
+      | BIC = OFV + n_estimated_iiv_omega_parameters * log(n_individals)
     Parameters
```
rikardn, line 567: I think there has to be an empty line before Parameters.

rikardn (approved):
Looks good. The documentation might not build as per comment above.

src/pharmpy/modeling/results.py
ZheHuang96: Ah, sorry I forgot to keep the empty line! I do it now!

PR #617, opened 2022-03-15
Use the same time varying cutoff for resmod models and best model

ZheHuang96:
Hi Rikard,

This PR is to fix an issue in `create_best_model `function when transforming time varying model. because the resmod dataset and nlme model dataset are different, I should use the tad column in resmod dataset rather than tad column in the original dataset when I transform the best model.

Best regards,
Zhe Huang


ZheHuang96:
Hi Rikard,
I am so sorry that I forgot you added a model copy line in the `create_dataset `function. Yes, you are right, we didn't add the tad in the original model dataset, instead, we add tad in the copy model dataset.

rikardn (approved):
Looks good now!

ZheHuang96:
Sorry I just see the comment! Thank you!

PR #621, opened 2022-03-16
Compare nlme model ofv difference with cutoff in each iteration of resmod

ZheHuang96:
Hi Rikard,

This PR is to compare the delta_ofv of the nlme model with the cutoff to decide whether we should do the next resmod iteration or not. I also add p_value here in case we need to change in the future, 0.05 is the default. 

Best regards,
Zhe Huang

### DIFF
--- a/src/pharmpy/modeling/amd.py
+++ b/src/pharmpy/modeling/amd.py
@@ -121,24 +121,34 @@ def _run_iiv(model, path):
 def _run_resmod(model, path):
     skip = []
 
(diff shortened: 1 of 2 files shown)


rikardn:
I am wondering if this could be put inside the resmod tool in some way. Perhaps the output model of resmod should only be an updated model in case the fit of the full model is significantly better.

ZheHuang96:
Ok, I can put the comparison inside the resmod tool :)

PR #627, opened 2022-03-16
Compare full models in resmod and avoid unnecessary iteration in amd

ZheHuang96:
Hi Rikard,

I change the commit message so I close the previous PR and open a new one. I use the same cutoff for CWRES model comparison and full model comparison. If there is no significant difference between the start model and the best model generated by resmod, the best model will be the start model and the selected_model_name will be called 'base' so that we can stop the iteration in amd when the model name is 'base'.

Best regards,
Zhe Huang


ZheHuang96:
But I am not so sure why the tox checks don't run.

rikardn (approved):
LGTM

PR #633, opened 2022-03-17
Configure pyright to use py310 virtual env generated by tox.

pharmpy-dev-123:


rikardn:
LGTM and I don't expect this to break the tests so I merge.

PR #634, opened 2022-03-18
Add summary_individuals attribute to ModelSearchResults

pharmpy-dev-123:
Implementation at https://github.com/pharmpy/pharmpy/blob/48674953f5bcdc649ccbd712c8617414cad6f39c/src/pharmpy/modeling/summarize_individuals.py#L10
Attribute definition at https://github.com/pharmpy/pharmpy/blob/48674953f5bcdc649ccbd712c8617414cad6f39c/src/pharmpy/tools/modelsearch/tool.py#L87
Includes one integration test at https://github.com/pharmpy/pharmpy/blob/48674953f5bcdc649ccbd712c8617414cad6f39c/tests/integration/test_modelsearch.py#L222

### DIFF
--- a/src/pharmpy/modeling/summarize_individuals.py
+++ b/src/pharmpy/modeling/summarize_individuals.py
@@ -1,0 +1,123 @@
+from typing import Callable, Dict, List, Union
+
+import numpy as np
+import pandas as pd
+
+from pharmpy.model import Model
+from pharmpy.modeling.ml import predict_influential_individuals, predict_outliers
+
+
+def summarize_individuals(models: List[Model]) -> Union[pd.DataFrame, None]:
+    """Creates a summary dataframe keyed by model-individual pairs for an input
+    list of models.
+
+    Parameters
+    ----------
+    models : List[Model]
+        Input models
+
+    Return
+    ------
+    pd.DataFrame | None
+        The summary as a dataframe
+
+    Examples
+    --------
+    >>> from pharmpy.modeling import *
+    >>> model = load_example_model("pheno")
+    >>> fit(model)
+    <Pharmpy model object pheno>
+    >>> results = run_tool(
+    ...     'modelsearch',
+    ...     model=model,
+    ...     mfl='ABSORPTION(ZO);PERIPHERALS([1, 2])',
+    ...     algorithm='reduced_stepwise'
+    ... ) # doctest: +SKIP
+    >>> summarize_individuals([results.start_model, *results.models]) # doctest: +SKIP
+
+    """
+
+    modelsDict = {model.name: model for model in models}
+
+    df = pd.concat(
+        map(
+            lambda model: groupedByIDAddColumnsOneModel(modelsDict, model),
+            models,
+        ),
+        keys=[model.name for model in models],
+        names=['model'],
+        axis=0,
+        verify_integrity=True,
+    )
+
+    return df
+
+
+def parent_model_name(model: Model) -> str:
+    return model.parent_model
+
+
+def model_name(model: Model) -> str:
+    return model.name
+
+
+def outliers_fda_func(df: pd.DataFrame) -> float:
+    # NOTE this returns a float because we will potentially concat this
+    # with NaNs
+    return float((abs(df) > 5).sum())
+
+
+def outliers_fda(model: Model) -> Union[pd.Series, float]:
+    residuals = model.modelfit_results.residuals
+    if residuals is None:
+        return np.nan
+    else:
+        groupedByID = residuals.groupby('ID')
+        return groupedByID['CWRES'].agg(outliers_fda_func)
+
+
+def _predicted(
+    predict: Callable[[Model], pd.DataFrame], model: Model, column: str
+) -> Union[pd.Series, float]:
+    try:
+        predicted = predict(model)
+    except TypeError:
+        # TODO remove this check as a TypeError should not be thrown by a
+        # predictor, but None should be returned instead
+        return np.nan
+    if predicted is None:
+        return np.nan
+    return predicted[column]
+
+
+def predicted_residual(model: Model) -> Union[pd.Series, float]:
+    return _predicted(predict_outliers, model, 'residual')
+
+
+def predicted_dofv(model: Model) -> Union[pd.Series, float]:
+    return _predicted(predict_influential_individuals, model, 'dofv')
+
+
+def ofv(model: Model) -> pd.Series:
+    return model.modelfit_results.individual_ofv
+
+
+def dofv(parent_model: Union[Model, None], candidate_model: Model) -> Union[pd.Series, float]:
+    return np.nan if parent_model is None else ofv(parent_model) - ofv(candidate_model)
+
+
+def groupedByIDAddColumnsOneModel(modelsDict: Dict[str, Model], model: Model) -> pd.DataFrame:
+    df = pd.DataFrame.from_dict(
+        {
+            'parent_model': parent_model_name(model),
+            'outliers_fda': outliers_fda(model),
+            'ofv': ofv(model),
+            'dofv': dofv(modelsDict.get(model.parent_model), model),
+            'predicted_dofv': predicted_dofv(model),
+            'predicted_residual': predicted_residual(model),
+        },
+        orient='columns',
+    )
+    df.reset_index(inplace=True)
+    df.set_index(['ID'], inplace=True)
+    return df
--- a/src/pharmpy/modeling/__init__.py
+++ b/src/pharmpy/modeling/__init__.py
@@ -159,6 +159,7 @@ from .results import (
     summarize_modelfit_results,
     write_results,
 )
+from .summarize_individuals import summarize_individuals
 from .units import get_unit_of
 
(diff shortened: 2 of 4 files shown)


src/pharmpy/modeling/summary_individuals.py
```diff
@@ -0,0 +1,108 @@
...
+def ofv(model: Model) -> pd.Series:
+    return model.modelfit_results.individual_ofv
+
+def groupedByIDAddColumnsOneModel(model: Model) -> pd.DataFrame:
+    df = pd.concat([
+        parent_model_name(model),
+        outliers_fda(model),
+        ofv(model),
+        predicted_dofv(model),
+        predicted_residual(model)
```
pharmpy-dev-123, line 104: This list could be parameterizable.

src/pharmpy/modeling/summary_individuals.py
```diff
@@ -0,0 +1,108 @@
...
+
+    df['diOFV'] = df.apply(lambda row: row['iOFV'] - df.loc[(row['parent_model'], row.name[1])]['iOFV'], axis=1)
+
+    return df
+
+def constant_series(model: Model, name: str, value) -> pd.Series:
+    index = model.dataset['ID'].unique()
+    series = pd.Series(np.repeat(value, len(index)), index=index, name=name)
+    series.index.rename('ID', inplace=True)
+    return series
```
pharmpy-dev-123, line 53: Maybe there is a better way?

src/pharmpy/modeling/summary_individuals.py
```diff
@@ -0,0 +1,108 @@
...
+    if results.start_model is None: return None
+
+    models = [results.start_model] if results.models is None else [results.start_model, *results.models]
+
+    df = pd.concat(map(
+        groupedByIDAddColumnsOneModel,
+        models,
+    ), keys=[model.name for model in models], names=['model'], axis=0, verify_integrity=True)
+
+    df['diOFV'] = df.apply(lambda row: row['iOFV'] - df.loc[(row['parent_model'], row.name[1])]['iOFV'], axis=1)
```
pharmpy-dev-123, line 45: Any idea how to improve readability?

src/pharmpy/modeling/summary_individuals.py
```diff
@@ -0,0 +1,108 @@
...
+    >>> model = load_example_model("pheno")
+    >>> fit(model)
+    >>> results = run_tool('modelsearch', model=model, mfl='ABSORPTION(ZO);PERIPHERALS([1, 2])', algorithm='reduced_stepwise')
+    >>> summary_individuals(results) # doctest: +SKIP
+
+    """
+
+    if results.start_model is None: return None
+
+    models = [results.start_model] if results.models is None else [results.start_model, *results.models]
```
pharmpy-dev-123, line 38: Maybe returning `None` is not the best way to handle missing data.

src/pharmpy/modeling/summary_individuals.py
```diff
@@ -0,0 +1,108 @@
...
+def parent_model_name(model: Model) -> pd.Series:
+    return constant_series(model, 'parent_model', model.parent_model)
+
+def model_name(model: Model) -> pd.Series:
+    return constant_series(model, 'model', model.name)
+
+def outliers_fda_func(df: pd.DataFrame) -> float:
+    # NOTE this returns a float because we will potentially concat this
+    # with NaNs
+    return float((abs(df) > 5).sum())
```
pharmpy-dev-123, line 64: `5` or the entire condition could be made parameterizable

src/pharmpy/modeling/summary_individuals.py
```diff
@@ -0,0 +1,108 @@
...
+import pandas as pd
+import numpy as np
+
+from pharmpy.tools.modelsearch.tool import ModelSearchResults
+from pharmpy.model import Model
+
+from pharmpy.modeling.ml import predict_outliers
+from pharmpy.modeling.ml import predict_influential_individuals
+
+def summary_individuals(results: ModelSearchResults) -> Union[pd.DataFrame,None]:
```
rikardn, line 12: We usually have a verb in the function names so perhaps `summarize_individuals`

src/pharmpy/modeling/summary_individuals.py
rikardn: Yes, you can use broadcasting of one single value into a column: `df['mycolumn'] = np.nan`

src/pharmpy/modeling/summary_individuals.py
```diff
@@ -0,0 +1,108 @@
...
+    """Creates a summary dataframe keyed by model-individual pairs for an input
+    ModelSearchResults object.
+
+    Parameters
+    ----------
+    results : ModelSearchResults
```
rikardn, line 18: If only a startmodel and all candidate models are used these should be the parameters instead of the ModelSearchResults. That way the function would be more general.

src/pharmpy/modeling/summary_individuals.py
rikardn: Something along the lines of `model.modelfit_results.individual_ofv - base_model.modelfit_results.individual_ofv` perhaps

src/pharmpy/modeling/summary_individuals.py
rikardn: We could leave as 5 for now

src/pharmpy/modeling/summary_individuals.py
rikardn: I think it is fine

src/pharmpy/modeling/summary_individuals.py
pharmpy-dev-123: There is also `pd.DataFrame({'a': [1,2,3], 'b': np.nan})` instead of `pd.concat`.

src/pharmpy/modeling/summarize_individuals.py
```diff
@@ -0,0 +1,139 @@
...
+def ofv(model: Model) -> pd.Series:
+    return model.modelfit_results.individual_ofv
+
+
+def dofv(modelA: Model, modelB: Model) -> pd.Series:
+    return ofv(modelA) - ofv(modelB)
+
+
+def groupedByIDAddColumnsOneModel(modelsDict: dict[str, Model], model: Model) -> pd.DataFrame:
+    df = pd.DataFrame.from_dict(
```
pharmpy-dev-123, line 126: I think we need to explicitly set the index here in case none of the columns are series.

src/pharmpy/modeling/summarize_individuals.py
```diff
@@ -0,0 +1,139 @@
...
+    >>> from pharmpy.modeling import *
+    >>> model = load_example_model("pheno")
+    >>> fit(model)
+    >>> results = run_tool(
+    >>>     'modelsearch',
+    >>>     model=model,
+    >>>     mfl='ABSORPTION(ZO);PERIPHERALS([1, 2])',
+    >>>     algorithm='reduced_stepwise'
+    >>> )
+    >>> summarize_individuals(results) # doctest: +SKIP
```
pharmpy-dev-123, line 36: This line need to be updated to reflect new API. Are we happy with a list of models as input? Should we check that there are no missing parents?

rikardn: I think a list is fine. If you want you could check for missing parents. In that case I case the diOFV couldn't be computed.

src/pharmpy/modeling/summarize_individuals.py
rikardn: I don't know if I understand this. A column in a dataframe is always a series.

pharmpy-dev-123: If all the values in the dictionary are constants the 'ID' column will not exist. Currently, I do not think it is possible it would happen.

src/pharmpy/modeling/summarize_individuals.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def ofv(model: Model) -> pd.Series:
+    return model.modelfit_results.individual_ofv
+
+
+def dofv(modelA: Model, modelB: Union[Model, None]) -> Union[pd.Series, float]:
+    if modelB is None:
+        return np.nan
+    return ofv(modelA) - ofv(modelB)
```
pharmpy-dev-123, line 107: Is it `child - parent` or `parent - child`?

rikardn:
There is a failing doctest. I think more `SKIP` is needed. You can run all doctests with `tox -e doctest`. I don't know how to run one single test.

pharmpy-dev-123:
> There is a failing doctest. I think more `SKIP` is needed. You can run all doctests with `tox -e doctest`. I don't know how to run one single test.

Indeed! Amended and force-pushed. I had to add `SKIP` because it was complaining about not finding the `nmfe` script. Maybe there is a better way?

src/pharmpy/modeling/summarize_individuals.py
pharmpy-dev-123: It is `parent - child`.

rikardn (approved):
LGTM

PR #659, opened 2022-03-31
[feature] `resume_tool(path)`

### LINKED ISSUE #742: Make modelfit `Results` serializable/deserializable

pharmpy-dev-123:
> @rikardn Ready to MERGE

  - [x] Rebase on #737.
  - [x] Rebase on #704 (I rebased and closed the original PR).
  - [x] Fixes #742.
  - [x] Use tool metadata to implement a `resume_tool(path)` helper function.
  - [x] Test `resume_tool(...)`
    - [x] Test it does not crash
    - [x] Test if returns same results as original call
  - [x] Minimize risks of failures given current imperfect model serialization by not running `run_tool` on serialized metadata (which is what `resume_tool` would do, and would be better to do in the future to ensure both do exactly the same thing). This is a first draft: it works, but it would work better and more reliably once  a fix for #743 (for instance #1211) is implemented.
  - [x] Fix bug with `run_amd`: because we rename the starting model to `input_model`, running multiple tools in sequence such as in `run_amd` causes some problems. See #1370.
  - [x] Do not change naming of models input to tools.
  - [x] Integration tests: https://github.com/pharmpy-dev-123/pharmpy/actions/workflows/integration.yml?query=branch%3Afeature-checkpointing-v0


tests/execute/test_database.py
```diff
@@ -101,3 +114,27 @@ def test_store_model(testdata):
...
+        create_joint_distribution(model_iiv)
+
+        assert model_start != model_iiv
+
+        db = LocalModelDirectoryDatabase("database")
+        db.store_model(model_iiv)
+
+        model_iiv_read = db.get_model(model_iiv.name)
+
+        assert model_iiv_read == model_iiv
```
pharmpy-dev-123, line 140: This does not pass yet.

rikardn:
Resolved conflicts and rebased. Should be ready to merge if everything passes

PR #663, opened 2022-04-01
Change the best model name and predecessors of task_results in resmod (fixes pharmpy#635)

ZheHuang96:
Hej Rikard,

I just finished the abstract and am waiting for review, so I think I can go back to my development tasks when I am waiting:) This commit is to solve the issue of comparing the full model results in resmod.

Best regards,
Zhe Huang


rikardn (approved):
LGTM

PR #664, opened 2022-04-01
[refactor] Model plugin run metadata v0

pharmpy-dev-123:
- [x] Store `stdout`, `stderr`, `returncode`, command line arguments, wrapper path, and other plugin-specific metadata
    - [x] NONMEM
    - [x] NLMIXR
  - [x] Store `results.json` and other plugin-generic metadata
    - [x] NONMEM
    - [x] NLMIXR
  - [x] ~Have a way to flag that pharmpy was not interrupted in the middle of database writes~ See #671
  - [x] ~Store model file (hash?) for cache invalidation purposes~ Out of scope for now?
  - [x] ~Store data file (hash?) for cache invalidation purposes~ This will be handled by the dataset database.

### DIFF
--- a/src/pharmpy/plugins/nlmixr/model.py
+++ b/src/pharmpy/plugins/nlmixr/model.py
@@ -1,3 +1,4 @@
+import json
 import os
 import subprocess
 import uuid
@@ -231,7 +232,8 @@ def execute_model(model):
     model = convert_model(model)
     path = Path.cwd() / f'nlmixr_run_{model.name}-{uuid.uuid1()}'
     model._path = path
-    path.mkdir(parents=True, exist_ok=True)
+    meta = path / '.pharmpy'
+    meta.mkdir(parents=True, exist_ok=True)
     write_csv(model, path=path)
 
     code = model.model_code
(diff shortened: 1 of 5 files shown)


src/pharmpy/plugins/nlmixr/model.py
```diff
@@ -253,9 +254,41 @@ def execute_model(model):
...
     newenv['R_LIBS_USERS'] = ''
     newenv['R_LIBS_SITE'] = ''
-    subprocess.run([rpath, path / (model.name + '.R')], env=newenv)
+
+    stdout = path / 'stdout'
+    stderr = path / 'stderr'
+
+    args = [rpath, path / (model.name + '.R')]
+
+    with open(stdout, "wb") as out, open(stderr, "wb") as err:
```
rikardn, line 263: Do we need to open these? I think that `sys.stdout` and `sys.stderr` are always open.

src/pharmpy/plugins/nonmem/run.py
```diff
@@ -40,18 +41,46 @@ def execute_model(model):
...
         os.chmod(path / 'cdwrapper', 0o744)
         cmd = str(path / 'cdwrapper')
-    subprocess.call(
-        [cmd], stdin=subprocess.DEVNULL, stderr=subprocess.DEVNULL, stdout=subprocess.DEVNULL
-    )
+
+    stdout = path / 'stdout'
+    stderr = path / 'stderr'
+
+    with open(stdout, "wb") as out, open(stderr, "wb") as err:
```
rikardn, line 48: Same comment as above here.

src/pharmpy/plugins/nonmem/run.py
```diff
@@ -40,18 +41,46 @@ def execute_model(model):
...
     database.store_local_file(model, (path / basepath).with_suffix('.cov'))
     database.store_local_file(model, (path / basepath).with_suffix('.cor'))
     database.store_local_file(model, (path / basepath).with_suffix('.coi'))
+
     for rec in model.control_stream.get_records('TABLE'):
         database.store_local_file(model, path / rec.path)
+
+    database.store_optional_local_file(model, stdout)
+    database.store_optional_local_file(model, stderr)
+    database.store_optional_json(model, plugin, 'nonmem.json')
```
rikardn, line 79: Do we need two json-files and two methods? Could `store_metadata` suffice?

src/pharmpy/workflows/model_database/baseclass.py
```diff
@@ -37,6 +37,34 @@ def store_local_file(self, model, path):
         """
         pass
 
+    @abstractmethod
+    def store_optional_local_file(self, model, path):
```
rikardn, line 41: What does "optional" in the name refer to? Could the `store_local_file` method have been used?

src/pharmpy/plugins/nlmixr/model.py
pharmpy-dev-123: These are not `sys.stdout`/`sys.stdin`. These are temporary file paths. Maybe naming is confusing?

src/pharmpy/workflows/model_database/baseclass.py
pharmpy-dev-123: All files stored by `*_optional_*` methods are not written to disk in the `LocalDirectoryDatabase` implementation. Could have a better naming.

src/pharmpy/plugins/nonmem/run.py
pharmpy-dev-123: This is to separate plugin-generic stuff from plugin-specific stuff. One thing that can be done to reduce the number of methods is to first write the JSON to a local file, then use the existing `store_optional_local_file`.

pharmpy-dev-123:
Ready for final review.

PR #671, opened 2022-04-04
[refactor] Database transactions v0

pharmpy-dev-123:
The goal is to have a generic way to do atomic reads and writes on the different databases.

**TODO**:
  - [ ] Sanity check we are not going to end up in a dead-lock with current usage.
    - @rikardn I think there are some issues where reading the `modelfit_results` attribute inside any of the database methods may cause locking recursion. We should be able to detect if this happens by using nonreentrant thread locks.

**DONE**:
  - [x] Make it work with processes.
    - [x] **UNIX** (using `fcntl`)
      - [x] `blocking`
      - [x] `shared`
    - [x] **Windows** (using `msvcrt`) 
      - [x] `blocking`: Make it block forever (so that it behaves like `fcntl`)
      - [x] `shared`: ~Do not "simulate" a shared acquire using an exclusive acquire.~ Good enough for now. Maybe not possible without additional libraries such as `win32file` (see for instance https://github.com/WoLpH/portalocker/blob/00cffcb1831208691d3bf6850df4768989a0bd4c/portalocker/portalocker.py). 
    - [x] Make sure `fcntl` and `msvcrt` create advisory non-persistent locks (see https://github.com/conan-io/conan/issues/6079). Difficult to find documentation on this, but @ZheHuang96 and I did some testing on Windows and the behavior we observed is expected (everything is automatically released on process exit, files can be opened even when a lock exists, in fact you need to open first and lock given the obtained `fd`).
  - [x] Make it work with threads. Perhaps using `threading.Lock` with some global fs-path-to-lock mapping.
  - [x] Allow many-read XOR single-write per-database locks. Using `LOCK_SH` but making sure multi-threading does not wrongfully upgrade an existing `LOCK_EX` lock. This is achieved by acquiring a lock at the thread-level (with some `threading.Condition` and counting magic), then acquiring the process-level lock.
  - [x] Make it clear it only works with existing file paths (file has to exist on both UNIX and Windows, path cannot be a directory on Windows)

Some links in the wild:
  - https://www.oreilly.com/library/view/python-cookbook/0596001673/ch04s25.html
  - https://github.com/harlowja/fasteners
  - https://github.com/WoLpH/portalocker/blob/00cffcb1831208691d3bf6850df4768989a0bd4c/portalocker/portalocker.py
  - https://docs.openstack.org/pylockfile/latest/

### DIFF
--- a/src/pharmpy/lock.py
+++ b/src/pharmpy/lock.py
@@ -1,0 +1,237 @@
+"""The lock module exposes a path_lock context manager function
+
+Other exported functions are implementation details subject to change.
+
+Caveats:
+    If you need to lock multiple paths you should have a total order for each
+    subset of paths that could be locked simultaneously by the same thread and call
+    path_lock multiple times respecting this total order. For instance via
+    resolved absolute path lexicographical order.
+
+    Currently, on Windows, at the process level, shared locks are "simulated" by
+    exclusive locks. Implementing true shared locks on this platform would require
+    depending on userland libraries. This would be necessary, for instance, in
+    situations where multiple processes need to lock shared resources and
+    communicate simultaneously about the state of those resources. Note that the
+    problem in this particular example can be circumvented at the cost of
+    efficiency, by simulating locked shared resources through multiple rounds of
+    exclusively-locked resources plus communication. I do not think our current
+    usage mandates the correct implementation of shared locks. It is enough that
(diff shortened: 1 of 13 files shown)


pharmpy-dev-123:
@rikardn This is ready for review.

PR #675, opened 2022-04-05
[feature] Expand tilde to user `$HOME` when handling user input paths

### LINKED ISSUE #629: Support ~ as $HOME in all paths
Create a wrapper function to convert str->Path

pharmpy-dev-123:
Fixes #629.

### DIFF
--- a/src/pharmpy/config.py
+++ b/src/pharmpy/config.py
@@ -62,7 +62,6 @@ class ConfigItem:
         return instance.__dict__.get(self.name, self.default)
 
     def __set__(self, instance, value):
-        # if self.cls != type(value):
         try:
             if isinstance(self.default, list) and isinstance(value, str):
                 value = [v.strip() for v in value.split(',')]
--- a/src/pharmpy/modeling/common.py
+++ b/src/pharmpy/modeling/common.py
@@ -11,6 +11,7 @@ import sympy
 from pharmpy import Model, Parameters, RandomVariables, config
 from pharmpy.modeling import split_joint_distribution
 from pharmpy.statements import Assignment, CompartmentalSystem
+from pharmpy.utils import normalize_user_given_path
 from pharmpy.workflows import default_model_database
 
 
@@ -38,6 +39,7 @@ def read_model(path):
     read_model_from_string : Read model from string
 
     """
+    path = normalize_user_given_path(path)
     model = Model.create_model(path)
     return model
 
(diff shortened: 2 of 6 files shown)


src/pharmpy/config.py
```diff
@@ -47,13 +47,14 @@ def read_configuration():
 
 
 class ConfigItem:
-    def __init__(self, default, description, cls=None):
+    def __init__(self, default, description, cls=None, transform=None):
```
pharmpy-dev-123, line 50: Collateral damage.

src/pharmpy/plugins/nonmem/__init__.py
```diff
@@ -38,7 +39,10 @@ class NONMEMConfiguration(config.Configuration):
         list,
     )
     default_nonmem_path = config.ConfigItem(
-        pathlib.Path(''), 'Full path to the default NONMEM installation directory', cls=pathlib.Path
+        pathlib.Path(''),
+        'Full path to the default NONMEM installation directory',
+        cls=pathlib.Path,
+        transform=ux.normalize_user_given_path,
```
pharmpy-dev-123, line 45: Not sure this belongs here.

src/pharmpy/workflows/execute.py
```diff
@@ -102,6 +103,8 @@ def split_common_options(d):
     other_options = dict()
     for key, value in d.items():
         if key in execute_options:
+            if key == 'path':
+                value = normalize_user_given_path(value)
```
pharmpy-dev-123, line 107: Not sure this belongs here either.

pharmpy-dev-123 (commented):
I left some comments on the first draft.

src/pharmpy/plugins/nonmem/__init__.py
rikardn: I think you don't need the extra transform argument here. `cls` could be set to any callable returning an object of the class.

src/pharmpy/workflows/execute.py
rikardn: Makes sense to have here. Good catch

rikardn (changes requested):

src/pharmpy/modeling/common.py
```diff
@@ -11,6 +11,7 @@
 from pharmpy import Model, Parameters, RandomVariables, config
 from pharmpy.modeling import split_joint_distribution
 from pharmpy.statements import Assignment, CompartmentalSystem
+from pharmpy.ux import normalize_user_given_path
```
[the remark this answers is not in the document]
pharmpy-dev-123: Ok

rikardn:
Thanks!

PR #706, opened 2022-04-13
[feature] Use standard datetime format in tool metadata

pharmpy-dev-123:
So that it is parseable while staying readable.


stellabelin:
I think this looks good! You can maybe remove the `fixme` about the time format!

pharmpy-dev-123:
> I think this looks good! You can maybe remove the `fixme` about the time format!

Done!

rikardn (approved):

PR #735, opened 2022-04-21
Generate correct dataset for start model in amd

ZheHuang96:
Hi Rikard,

This PR is the solution for today's issue. I feel like it may be better to commit now rather than after all issues are solved:)

Best regards,
Zhe Huang


rikardn:
Good, it is always better to fix issues one at a time.

PR #736, opened 2022-04-22
[fix] Fix dataset datainfo datatype synchronization

### LINKED ISSUE #703: Identical datasets saved in .datasets in the run_iiv results
I run run_iiv using v0.63.2 and found each candidate model has its own dataset saved in the folder .datasets and they are identical. I thought the idea is to save the datasets that have been changed.

pharmpy-dev-123:
This fixes some issues regarding dataset comparisons happening when running the resmod tool. Might fix #703.

### DIFF
--- a/src/pharmpy/datainfo.py
+++ b/src/pharmpy/datainfo.py
@@ -79,6 +79,10 @@ class ColumnInfo:
         'subject identifier',
     ]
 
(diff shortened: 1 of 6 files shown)


rikardn:
This looks fine. I'll merge and then think about how the added methods fit into the overall API

PR #737, opened 2022-04-22
[feature] Add resume flag to `run_tool`

pharmpy-dev-123:
This allows to run a tool using an existing tool database at `path` with `run_tool(..., path=path, resume=True)`. Currently, the goal is only to support usage where `run_tool` is called twice with exactly the same arguments. If this constraint is met, the results should be correct (although perhaps not identical, we have an issue with the bootstrap seed not being identical). For tool runs that had already executed for a significant amount of time, it should be able to save time by recycling existing NONMEM/NLMIXR output.

More complex usage will eventually be covered later.


tests/integration/test_resume.py
```diff
@@ -0,0 +1,131 @@
...
+        (
+            'ABSORPTION([ZO,SEQ-ZO-FO]);PERIPHERALS(1)',
+            7,
+            'modelsearch_candidate6',
+            'modelsearch_candidate3',
+        ),
+        ('LAGTIME();TRANSITS(1)', 2, 'modelsearch_candidate2', 'mox2'),
+        ('ABSORPTION(ZO);TRANSITS(3, *)', 3, 'modelsearch_candidate2', 'mox2'),
+    ],
+)
```
pharmpy-dev-123, line 94: Maybe this is a bit exagerated?

src/pharmpy/workflows/model_database/local_directory.py
```diff
@@ -145,15 +151,37 @@ def retrieve_file(self, name, filename):
...
+        # model.modelfit_results = self.get_modelfit_results(name)
+        # except:
         model.read_modelfit_results()
         return model
 
+    def get_modelfit_results(self, name):
+        # FIXME The following does not work because deserialization of modelfit
+        # results is not generic enough.
+        # path = self.path / name / DIRECTORY_PHARMPY_METADATA / FILE_MODELFIT_RESULTS
+        # return read_results(path)
```
pharmpy-dev-123, line 182: I think this can go away. We will take care of this later.

src/pharmpy/workflows/model_database/local_directory.py
```diff
@@ -145,15 +151,37 @@ def retrieve_file(self, name, filename):
...
+            except FileNotFoundError as e:
+                errors.append(e)
+                pass
+        else:
+            raise FileNotFoundError(errors)
+
         model.database = self
+        # try:
+        # model.modelfit_results = self.get_modelfit_results(name)
+        # except:
```
pharmpy-dev-123, line 174: This only makes sense once `results.json` deserialization works.

src/pharmpy/workflows/model_database/local_directory.py
```diff
@@ -145,15 +151,37 @@ def retrieve_file(self, name, filename):
             raise FileNotFoundError(f"Cannot retrieve {filename} for {name}")
 
     def get_model(self, name):
-        filename = name + self.file_extension
-        path = self.path / name / filename
+        extensions = ['.ctl', '.mod']
         from pharmpy.model import Model
 
-        model = Model.create_model(path)
```
pharmpy-dev-123, line 152: Do I understand correctly that this only worked with NONMEM models before anyway?

src/pharmpy/workflows/model_database/local_directory.py
```diff
@@ -7,6 +7,9 @@
 
 from .baseclass import ModelDatabase
 
+# from pharmpy.results import read_results
+
+
```
pharmpy-dev-123, line 12: This can go away. See below.

src/pharmpy/tools/modelfit/tool.py
```diff
@@ -53,3 +43,50 @@ def post_process_results(*models):
...
+            if db_model and model.has_same_dataset_as(db_model):
+                # NOTE Inputs are identical so we can reuse the results
+                model.modelfit_results = db_results
+                return model
+
+        # NOTE Fallback to executing the model
+        execute_model = get_execute_model(tool)
+        return execute_model(model)
+
+    return task
```
pharmpy-dev-123, line 76: Maybe this could be split into a separate commit?

src/pharmpy/modeling/run.py
```diff
@@ -51,7 +51,12 @@ def fit(models, tool=None):
             db_model = read_model_from_database(model.name, database=model.database)
         except (KeyError, AttributeError):
             db_model = None
-        if db_model and db_model.modelfit_results is not None and db_model == model:
+        if (
+            db_model
+            and db_model.modelfit_results is not None
+            and db_model == model
+            and model.has_same_dataset_as(db_model)
```
pharmpy-dev-123, line 58: This now forces comparing model datasets when calling `fit(...)`.

src/pharmpy/model.py
```diff
@@ -311,6 +311,19 @@ def parent_model(self):
...
+            if other.dataset is None:
+                return True
+            else:
+                return False
+
+        if other.dataset is None:
+            return False
+
+        # NOTE rely on duck-typing here (?)
+        return self.dataset.equals(other.dataset)
```
pharmpy-dev-123, line 325: `(self.dataset == other.dataset).all().all()` would skip comparing `dtype`.

pharmpy-dev-123 (commented):
@rikardn Some comments.

rikardn (approved):
LGTM. Missing docstring for has_same_dataset_as. I can add it myself in separate commit.

PR #747, opened 2022-04-25
Amd subtools

ZheHuang96:
Hi Rikard,

According to the discussion today, I changed some options wich related to` run_amd` function and make this PR. About the separator problem, I still need to fix some problem for the integration test. I will re-looked at it tomorrow.

Best regards,
Zhe Huang

### DIFF
--- a/tests/modeling/test_modeling.py
+++ b/tests/modeling/test_modeling.py
@@ -1938,11 +1938,6 @@ def test_zo_abs_on_nl_elim(testdata):
             'CL = TVCL*EXP(ETAB1)\nV = TVV*EXP(ETAB2)',
         ),
         (
-            ['eta(1)'],
-            'ETAB1 = (EXP(ETA(1))**THETA(4) - 1)/THETA(4)',
-            'CL = TVCL*EXP(ETAB1)\nV=TVV*EXP(ETA(2))',
-        ),
-        (
             'ETA(1)',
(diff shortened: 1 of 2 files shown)


ZheHuang96:
Hi Rikard, 
The tox check failed when test `transform_etas_boxcox` with eta commit In this case, we have a `eta(1)` and I guess we need to keep it as uppercase?

tests/modeling/test_modeling.py
```diff
@@ -1937,11 +1937,6 @@ def test_zo_abs_on_nl_elim(testdata):
             'ETAB2 = (EXP(ETA(2))**THETA(5) - 1)/THETA(5)',
             'CL = TVCL*EXP(ETAB1)\nV = TVV*EXP(ETAB2)',
         ),
-        (
```
rikardn, line 1940: Couldn't this stay with 'ETA(1)' instead of 'eta(1)' ?

rikardn: True, perhaps a too small difference. You don't have to change it.

rikardn (approved):
I think this looks good.

tests/modeling/test_modeling.py
ZheHuang96: Ok, thank you!

ZheHuang96:
Thank you so much:)

PR #761, opened 2022-04-29
Add separator in datainfo file and read in dataset correctly

ZheHuang96:
Hi Rikard,

Here is the commit for removing the default separator,  @pharmpy-dev-123 and I found that we should put double backslash before the` s+` in the datainfo file for the white space separator, only one backslash doesn't work. 

Best regard,
Zhe Huang

### DIFF
--- a/src/pharmpy/modeling/example_models/pheno.datainfo
+++ b/src/pharmpy/modeling/example_models/pheno.datainfo
@@ -5,4 +5,4 @@
 	{ "name": "WGT", "type": "covariate", "scale": "ratio", "unit": "kg", "descriptor": "body weight" },
 	{ "name": "APGR", "type": "covariate", "scale": "ordinal", "categories": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10] },
 	{ "name": "DV", "type": "dv", "scale": "ratio", "unit": "mg/l", "descriptor": "plasma concentration" }
-]}
+	], "separator":"\\s+"}
--- a/src/pharmpy/datainfo.py
+++ b/src/pharmpy/datainfo.py
@@ -640,6 +640,9 @@ class DataInfo(MutableSequence):
         if path:
             path = Path(path)
         di.path = path
+        separator = d.get('separator', None)
+        if separator:
+            di.separator = separator
         return di
 
     @staticmethod
--- a/src/pharmpy/modeling/data.py
+++ b/src/pharmpy/modeling/data.py
@@ -1442,15 +1442,12 @@ def read_dataset_from_datainfo(datainfo: Union[DataInfo, Path, str]) -> pd.DataF
     """
     if not isinstance(datainfo, DataInfo):
         datainfo = DataInfo.read_json(datainfo)
-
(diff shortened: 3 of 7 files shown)


rikardn:
The cause of the fail in the tests have been fixed on master. Could you rebase and push again?

src/pharmpy/modeling/example_models/pheno.datainfo
```diff
@@ -5,4 +5,4 @@
 	{ "name": "WGT", "type": "covariate", "scale": "ratio", "unit": "kg", "descriptor": "body weight" },
 	{ "name": "APGR", "type": "covariate", "scale": "ordinal", "categories": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10] },
 	{ "name": "DV", "type": "dv", "scale": "ratio", "unit": "mg/l", "descriptor": "plasma concentration" }
-]}
+	], "separator":"\\s+"}
```
rikardn, line 8: Could we figure out a way to avoid having to have double backslashes here?

ZheHuang96: Hi Rikard, 

It does not very user-friendly to have double backslashes. I tried to explore it last week but I didn't find any useful results. When we read in the string of a JSON file, "\s+" is illegal and it is not a special character like "\n", 
```
Traceback (most recent call last):

  Input In [19] in <cell line: 1>
    data = json.load(json_file)

  File ~\Anaconda3\lib\json\__init__.py:293 in load
    return loads(fp.read(),

  File ~\Anaconda3\lib\json\__init__.py:357 in loads
    return _default_decoder.decode(s)

  File ~\Anaconda3\lib\json\decoder.py:337 in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())

  File ~\Anaconda3\lib\json\decoder.py:353 in raw_decode
    obj, end = self.scan_once(s, idx)

JSONDecodeError: Invalid \escape
```
we have to use two backslashes, but after reading in, the separator only has one backslash which we actually want. 
```
with open('C:\\Users\\zhehu311\\Desktop\\debug\\amd\\separator.json') as json_file:
    data = json.load(json_file)
    print(data)
{'separator': '\\s+'}

print(data['separator'])
\s+
```
What I can think about is, that maybe we can add a functionality to detect and separate the '\s+' and ',' in pharmpy so that users don't need to write it in the JSON file. May I ask your idea about it?

Best regards,
Zhe Huang

pharmpy-dev-123: Yep escaping backslashes is mandatory with JSON. As @ZheHuang96 suggests we can hardcode a nice default value, or we could use our own encoding that does not use backslashes, but we have to define the bijection between tht encoding and the one understood by `pd.read_csv`.

rikardn: Ok, I didn't know. Turns out that JSON also has some quirks. Let us don't change this then.

rikardn (approved):

PR #763, opened 2022-04-29
[docs] Scroll x-axis overflow of all jupyter outputs

pharmpy-dev-123:
This fixes UI problems when tables are large.


rikardn:
Nice!

PR #764, opened 2022-04-29
Allometry tool

ZheHuang96:
Hi Rikard,

Here is a very basic version of allometry tool, maybe we can add more features to this tool. May I ask your idea about it?

Best regards,
Zhe Huang


rikardn:
Thanks! This is a good first version. We currently don't know exactly what more it should do.

PR #771, opened 2022-05-02
[fix] Make `add_iov` and `remove_iov` work with case statements

pharmpy-dev-123:
> :warning: Needs squash or commit editing

@rikardn Ready for review.

- [x] Fix AST corruption when calling `add_iov` then `remove_iov`.
- [x] Add some unit tests to catch regressions.


rikardn (approved):
Thanks. This is looking good. Do you want to squash or change the commit messages before merging?

PR #791, opened 2022-05-04
[docs] Automation of R examples for web documentation

### LINKED ISSUE #774: Automation of R examples for web documentation
For example tabs for Python, R and CMD-line. Possibility to execute Python code via jupyter-sphinx. Some implementation exists in pharmr and should be moved.

pharmpy-dev-123:
- [x] Generalize `jupyter-execute` to prepare for automated Python-to-R transpiling
  - [x] Automate Python-to-R transpiling
  - [x] Tweak Python-to-R transpiling
  - [x] Update all uses of `.. code-block:: python` and `.. jupyter-execute::`
  - [x] ~Make API docs code examples use transpilation directive (?) See https://github.com/astropy/sphinx-automodapi/issues/151~ Skip for now since we have dedicated R API docs anyway.
  - [x] Fixes #774

### DIFF
--- a/docs/_ext/conversion.py
+++ b/docs/_ext/conversion.py
@@ -1,6 +1,6 @@
-from inspect import getdoc
 import re
-
+from inspect import getdoc
+from typing import List
 
 def create_r_doc(func):
     doc = getdoc(func)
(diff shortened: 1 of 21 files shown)


docs/_ext/conversion.py
```diff
@@ -73,57 +73,76 @@ def create_r_returns(doc_list):
...
+    # must_load_library = any(map(is_import_line, lines))
+    filtered_lines = [line for line in lines if not(is_import_line(line))]
+
+    transpiled_lines = map(
+        lambda line: line_transpile_py_to_r(re.sub(pattern_start, '', line)),
+        filtered_lines,
+    )
+
+    return transpiled_lines
+    # return ['library(pharmr)', *transpiled_lines] if must_load_library else list(transpiled_lines)
```
pharmpy-dev-123, line 132: We remove all `import` statements. We assume an implicit `library(pharmr)` has already been run.

docs/_ext/conversion.py
```diff
@@ -73,57 +73,76 @@ def create_r_returns(doc_list):
 
     return doc_str
 
+pattern_start = re.compile(r'>>> |^\.\.\. ')
+# pattern_methods = re.compile(r'([A-Za-z_]\d*)\.([A-Za-z_]\d*)')
+pattern_methods = re.compile(r'([A-Za-z_]\d*)\.([A-Za-z_]\d*)(?=(?:[^"]|"[^"]*")*$)(?=(?:[^\']|\'[^\']*\')*$)')
```
pharmpy-dev-123, line 78: This attempts to not replace `.`s that are inside quotes.

docs/_ext/pharmpy_snippet.py
```diff
@@ -0,0 +1,199 @@
...
+
+        # for idx, line in enumerate(new_content):
+            # self.content.data.insert(idx, line)
+            # self.content.items.insert(idx, (None, idx))
+
+        # del self.content.data[len(new_content):]
+        # del self.content.items[len(new_content):]
+
+        # node = nodes.container()
+        # self.state.nested_parse(self.content, self.content_offset, node)
```
pharmpy-dev-123, line 50: This can go to the garbage.

docs/_ext/pharmpy_snippet.py
```diff
@@ -0,0 +1,199 @@
...
+        "linenos": directives.flag,
+        "lineno-start": directives.nonnegative_int,
+        "emphasize-lines": directives.unchanged_required,
+        "raises": csv_option,
+        "stderr": directives.flag,
+    }
+
+    def run(self):
+        # if 'hide-code' in self.options:
+            # raise ValueError('pharmpy-execute does not allow to hide code, use jupyter-execute instead')
```
pharmpy-dev-123, line 71: I thought it was a good idea to prevent `hide-code` to be used with `pharmpy-execute`, but it actually simplifies this PR to allow it, and it is also more general should we ever switch from jupyter to something else.

docs/_ext/pharmpy_snippet.py
```diff
@@ -0,0 +1,199 @@
...
+                *self._output(),
+                '',
+                *self._input(),
+            ]
+
+        return [
+            *self._input(),
+            '',
+            *self._output(),
+        ]
```
pharmpy-dev-123, line 101: We always have to return the output lines even when `hide-output` is set, otherwise the code will not be executed.

docs/_ext/pharmpy_snippet.py
```diff
@@ -0,0 +1,199 @@
...
+            ]),
+        ]
+
+    def _code_option_lines(self):
+        if 'emphasize_lines' in self.options:
+            yield f':emphasize_lines:{self.options.get("emphasize_lines")}'
+        if "linenos" in self.options:
+            yield ':linenos:'
+        if "lineno-start" in self.options:
+            yield f':lineno-start:{self.options.get("lineno-start")}'
```
pharmpy-dev-123, line 126: `emphasize_lines` and `lineno-start` may not work well with tranpsiled code, but that could be fixed later.

docs/_ext/pharmpy_snippet.py
```diff
@@ -0,0 +1,199 @@
...
+    def _jupyter_option_lines(self):
+        yield ':hide-code:'
+        if "hide-output" in self.options:
+            yield ':hide-output:'
+        if 'raise' in self.options:
+            yield f':raises:{",".join(self.options.get("raises"))}'
+        if "stderr" in self.options:
+            yield ':stderr:'
+
+class PharmpyCode(RecursiveDirective):
```
pharmpy-dev-123, line 147: Maybe there would be a way to share more code between `PharmpyCode` and `PharmpyExecute`.

stellabelin (commented):
Looks good! Answered some comments

docs/_ext/pharmpy_snippet.py
pharmpy-dev-123: Indeed.

docs/_ext/pharmpy_snippet.py
pharmpy-dev-123: Yes!

docs/_ext/conversion.py
pharmpy-dev-123: OK, but most examples in the docs reuse variables from previous examples. Or at least they are run in such a way that it is possible to do so. I see how it makes sense for doctests in the API docs, but not in the general documentation.

docs/_ext/conversion.py
pharmpy-dev-123: I do not understand what you mean? The regular expression logic essentially decides the parity of how many quotes are to the right of the dot. So it could be replaced by ad-hoc code.

docs/_ext/conversion.py
pharmpy-dev-123: Agree that if only `pharmpy.modeling` is available in `pharmr` there should be a note about it.

stellabelin: Okay, agree! I guess it can be said if you're trying pharmr examples you need to load pharmr, in Python it's more relevant with imports since you can import specific functions. Yes, in the R-documentation there is a note about it

docs/_ext/conversion.py
stellabelin: Sorry, you're right, thought about it wrong. Then this looks good!

docs/_ext/pharmpy_snippet.py
pharmpy-dev-123: Working on it.

pharmpy-dev-123: See https://github.com/pharmpy/pharmpy/pull/791/commits/08f529609f0ff4769e7c4a9426f1479e836b898c. What do you think?

stellabelin: I think it looks good, but is there an advantage to have three classes (`PharmpyAbstractCodeDirective`, `PharmpyExecute`, `PharmpyCode`) instead of having `PharmpyCode` and `PharmpyExecute(PharmpyCode)`? Are there any instances of `PharmpyAbstractCodeDirective`?

pharmpy-dev-123: I think I just wanted to avoid overloading concrete implementations of methods. So I inherit from a common abstract ancestor.

PR #793, opened 2022-05-05
Find correct location of model parameter in resmod result

ZheHuang96:
Hi Rikard,

This PR is to find the correct location of the model parameters from the result dataframe during the iterations.

Best regards,
Zhe Huang


rikardn (approved):
Good catch!

PR #797, opened 2022-05-06
Resmod Iteration

ZheHuang96:
Hi Rikard,

This PR is to fix the wrong start model for each iteration (it always used the first start model as the input model for each iteration so that it can not really do the iterations and generated the wrong best model code) and remove unnecessary iterations in amd. Now the resmod tool can work again:)

Best regards,
Zhe Huang

### DIFF
--- a/src/pharmpy/modeling/amd.py
+++ b/src/pharmpy/modeling/amd.py
@@ -153,29 +153,7 @@ def _run_iiv(model, path):
 
 
 def _run_resmod(model, path):
-    skip = []
-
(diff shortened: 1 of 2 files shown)


ZheHuang96:
The next step is to run amd in full scope(SIRAC) and change the test data for resmod tool. I found the moxo mod in our test data but I think maybe I also need a datainfo for moxo? I will test amd first then go back to the resmod test.

rikardn (approved):

PR #800, opened 2022-05-09
[refactor] Simplified and more lightweight implementation of CodeRecord#statements

pharmpy-dev-123:
> :warning: Needs commit editing. @rikardn Ready for review.

In this PR we use children index ranges to locate statement nodes in the AST. It is lighter on reference counting and avoids the linear scan to align the children array with the statement nodes array.


rikardn (approved):
Thanks for this cleanup.

PR #805, opened 2022-05-10
[fix] Avoid lazy parsing concurrency issues in NONMEM `Record`

pharmpy-dev-123:
> @rikardn Ready for merge.


src/pharmpy/plugins/nonmem/records/record.py
```diff
@@ -10,18 +10,13 @@ class Record:
     def __init__(self, content, parser_class):
         self._content = content
         self._parser_class = parser_class
-        self._root = None
+        self._root = self._parser_class(self._content).root
```
pharmpy-dev-123, line 13: Is it really necessary to keep a reference to `content` and `parser_class`? These will not be reused.

rikardn: Not if we are parsing directly

pharmpy-dev-123: OK removing those then.

rikardn (approved):
This looks good. We might need to revisit sooner than we think though. The main problem with always parsing might not the performance as I first thought but that fact that our grammars might not be able to parse all records in the wild. In some cases all records are not needed by Pharmpy but now they still need to be parsed potentially crashing the parser unnecessarily for more models.

PR #806, opened 2022-05-10
Add dummy statements for empty line node in start model

ZheHuang96:
Hi Rikard,

This PR is to make resmod part in amd run:)

Best regards,
Zhe Huang


rikardn (approved):
Thanks, this makes sense.

PR #821, opened 2022-05-17
[feature] IOV-search tool

pharmpy-dev-123:
> @rikardn Ready to merge.

  - [x] Rebase on #824 
  - [x] Discuss TODO points in `iovsearch/tool`:
    - [x] Default value for keyword argument `distribution`
    - [x] Allow passing an explicit list of ETAs/parameters to forward to the call of `add_iov`
    - [x] ~Should we exclude already present IOVs (both when adding IOVs, when removing IOVs, and when removing corresponding IIVs)?~ We assume there are no existing IOVs for now.
    - [x] ~Should we allow to remove all added IOVs (e.g. to test new initial estimates)?~ No that does not make sense.
  - [x] Write integration tests that cover all branches of the tool.

### DIFF
--- a/src/pharmpy/tools/common.py
+++ b/src/pharmpy/tools/common.py
@@ -69,7 +69,7 @@ def summarize_tool(
     rankfunc_name,
     cutoff,
     bic_type='mixed',
-):
+) -> pd.DataFrame:
     models_all = [start_model] + models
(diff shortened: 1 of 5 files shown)


rikardn (approved):
Looks good. You can go ahead and squash.

pharmpy-dev-123:
Commits have been squashed. Ready to merge.

PR #824, opened 2022-05-18
[feature/fix] Keep IIV random variables covariance structure when adding IOV

pharmpy-dev-123:
> @rikardn READY TO MERGE

  - [x] Naming? In particular `distribution="mixed"` could perhaps be `distribution="explicit"` or `distribution="given"`.
  - [x] Test all raising branches of `add_iov`.
  - [x] ~Currently, this PR removes the possibility of passing the same `ETA` "twice". Not only when passing exactly the same string as parameter, but also when passing both the ETA name and the corresponding parameter name. If both get resolved to the same ETA object, it will raise. This breaks some `add_iov` unit tests that I removed. Is there any reason to keep this "feature"?~ Yes we should keep this feature: an ETA could be on several parameters, so we want to handle that. We do not have to detect whether a parameter has several ETAs and simplify, for now. We still raise if the same ETA is given in disjoint distributions.
  - [x] Handle calling `add_iov` serveral times by using `pharmpy.modeling.expressions.create_symbol` to generate unique parameter names.
  - [x] Test `remove_iov` on IOVs with joint distributions.


src/pharmpy/modeling/eta_additions.py
```diff
@@ -133,6 +138,11 @@ def add_iov(model, occ, list_of_parameters=None, eta_names=None):
     eta_names : str, list
         Custom names of new etas. Must be equal to the number of input etas times the number of
         categories for occasion.
+    distribution : str
+        The distribution that should be used for the new etas. Options are
+        'disjoint' for disjoint normal distributions, 'joint' for joint normal
+        distribution, 'mixed' for an explicit mix of joint and disjoint
```
pharmpy-dev-123, line 144: Is `'mixed'` really a good name? I think I was influenced by some unrelated use of `'mixed'` in existing code.

src/pharmpy/modeling/eta_additions.py
```diff
@@ -156,44 +166,163 @@ def add_iov(model, occ, list_of_parameters=None, eta_names=None):
...
+            )
+        ):
+            raise ValueError('not all parameters are strings')
+
+        etas = [_get_etas(model, grp, include_symbols=True) for grp in list_of_parameters]
+
+        for dist, grp in zip(etas, list_of_parameters):
+            assert len(dist) <= len(grp)
+            if len(dist) != len(grp):
+                raise ValueError(f'Some ETA in {dist} was given twice.')
```
pharmpy-dev-123, line 215: Do we want to raise or silently support passing the same ETA twice?

src/pharmpy/modeling/eta_additions.py
```diff
@@ -156,44 +166,163 @@ def add_iov(model, occ, list_of_parameters=None, eta_names=None):
...
+    def eta_name(i, k):
+        return eta_names[(i - 1) * len(categories) + k - 1] if eta_names else f'ETA_IOV_{i}{k}'
 
-            values += [S(eta_new.name)]
-            conditions += [Eq(cat, S(occ))]
+    def omega_iov_name(i, j):
+        return f'OMEGA_IOV_{i}' if i == j else f'OMEGA_IOV_{i}_{j}'
+
+    def iov_name(i):
+        return f'IOV_{i}'
```
pharmpy-dev-123, line 257: This is bad in particular because we cannot run `add_iov` twice on the same `model`.

src/pharmpy/modeling/eta_additions.py
```diff
@@ -156,44 +166,163 @@ def add_iov(model, occ, list_of_parameters=None, eta_names=None):
...
+
+
+def _add_iov_mixed(model, occ, etas, categories, iov_name, eta_name, omega_iov_name):
+
+    ordered_etas = list(chain.from_iterable(etas))
+
+    eta, count = next(iter(Counter(ordered_etas).most_common()))
+
+    if count >= 2:
+        raise ValueError(f'{eta} was given twice.')
```
pharmpy-dev-123, line 275: Same comment as above: should we silently support giving the same ETA twice?

src/pharmpy/modeling/eta_additions.py
```diff
@@ -203,12 +332,45 @@ def add_iov(model, occ, list_of_parameters=None, eta_names=None):
...
+def _add_iov_etas_joint(rvs, pset, etas, indices, categories, omega_iov_name, eta_name):
+
+    _add_iov_declare_diagonal_omegas(rvs, pset, etas, indices, omega_iov_name)
+
+    # NOTE Declare off-diagonal OMEGAs
+    for i, j in combinations(indices, r=2):
+        omega_iov = S(omega_iov_name(i, j))
+        omega_iiv = rvs.get_covariance(etas[i - 1], etas[j - 1])
+        init = pset[omega_iiv].init * 0.1 if omega_iiv != 0 and omega_iiv in pset else 0.001
+        pset.append(Parameter(str(omega_iov), init=init))
```
pharmpy-dev-123, line 357: If `get_covariance(i,i)` was equivalent to `get_variance(i)` we could have a single implementation for diagonal and off-diagonal.

src/pharmpy/plugins/nonmem/update.py
```diff
@@ -199,7 +199,7 @@ def create_omega_single(model, rv, eta_number, record_number, comment_dict):
 
     if rv.level == 'IOV':
         param_str = f'${record_type}  BLOCK(1)'
-        first_iov = [iov for iov in rvs.iov if iov.parameter_names == rv.parameter_names][0]
+        first_iov = next(filter(lambda iov: iov.parameter_names == rv.parameter_names, rvs.iov))
```
pharmpy-dev-123, line 202: Bike-shedding?

src/pharmpy/plugins/nonmem/update.py
```diff
@@ -221,23 +221,32 @@ def create_omega_single(model, rv, eta_number, record_number, comment_dict):
...
-            if not re.match(r'OMEGA\(\d+,\d+\)', omega.name):
-                param_str += f'\t; {omega.name}'
-            elif comment_dict and omega.name in comment_dict:
-                param_str += f'\t; {comment_dict[omega.name]}'
-
-            param_str += '\n'
+    if rv.level == 'IOV' and rv != next(
+        filter(lambda iov: iov.parameter_names == rv.parameter_names, model.random_variables.iov)
+    ):
+        param_str += ' SAME\n'
```
pharmpy-dev-123, line 231: This was needed to handle outputting joint distributions for IOVs.

tests/testdata/nonmem/pheno_block.mod
```diff
@@ -1,6 +1,6 @@
 $PROBLEM PHENOBARB SIMPLE MODEL
 $DATA pheno.dta IGNORE=@
-$INPUT ID TIME AMT WGT APGR DV
+$INPUT ID TIME AMT WGT APGR DV FA1
```
pharmpy-dev-123, line 3: I need this for `same-as-iiv` tests that implicitly yield a mix of disjoint and joint distributions.

pharmpy-dev-123 (commented):
Some comments.

PR #858, opened 2022-06-02
AMD results first version

ZheHuang96:
Hi Rikard, 

This PR is to concatenate the results of modelsearch, iivsearch and resmod. I didn't write the AMD test because I was not sure it will still lead to some issues or not. But I think we need to check what the tables look like, I printed them, and here are the results:  
Screenshot from 2022-06-02 16-57-50
Screenshot from 2022-06-02 16-57-57
The way I concatenated the subtables is not very decent, it looks foolish...I did many tries today and this is the only way that works for me. I guess there might be a clear way which I didn't find online. I reset each subtable's index before concatenation because iivsearch has tuple multiIndex: step & model and` pd.concat(df, keys=keys)`  doesn't work in this situation. I also want to split the tuple and make the step a column so that I can fill numbers in it for other subtools. But after reset the index, it will generate a simple index like: 0,1,2... When I set index with multiIndex: tool, step and model, it will become a column so I have to drop it in the end:(

Best regards,
Zhe Huang

### DIFF
--- a/src/pharmpy/modeling/amd.py
+++ b/src/pharmpy/modeling/amd.py
@@ -1,5 +1,7 @@
 from functools import partial
 
+import pandas as pd
+
 import pharmpy.tools.scm as scm
 from pharmpy import plugins
 from pharmpy.results import Results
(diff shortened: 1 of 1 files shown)


rikardn:
Thanks

PR #869, opened 2022-06-08
Add allometric variable and combine iov tool in amd

ZheHuang96:
Hi Rikard,

This PR is based on our discussion this morning. There are some things that can be improved in the future.

- I think the allometry tool is under construction so I didn't combine allometry results in amd.

- We may need to implement the occasion in datainfo(maybe add a descriptor?) so that when the occasion is none, we can try to find a column for iovsearch which is similar to our strategy for allometry.

 Best regard,
 Zhe Huang


ZheHuang96:
Great, then I can combine them together today:)

PR #888, opened 2022-06-14
Insert covsearch tool and combine covsearch and allometry results in amd

ZheHuang96:
Hi Rikard,

This PR is to combine covsearch tool by using searching space and add results in AMD results class. 

I will set up the amd test after this PR

Best regards, 
Zhe Huang


ZheHuang96:
Hi Rikard, 

I tested the amd function with my own data and simulated mox normal data. I use the reduced search space for modelsearch to save time.

 My own data:

- modelsearch, iivsearch, resmod, allometry can pass
- iovsearch is not triggered because of no occasion in the dataset
- covsearch has an issue and I have asked Aurelien for help.

simulated mox normal data:
There is an issue when convert start_model to nonmem model.
```
tests/integration/test_amd.py:7 (test_amd)
test_amd.py:14: in test_amd
    res = run_amd(
../../src/pharmpy/tools/run_amd.py:102: in run_amd
    model = convert_model(model, 'nonmem')  # FIXME: Workaround for results retrieval system
../../src/pharmpy/modeling/common.py:238: in convert_model
    new = nonmem.convert_model(model)
../../src/pharmpy/plugins/nonmem/model.py:88: in convert_model
    nm_model.update_source()
../../src/pharmpy/plugins/nonmem/model.py:221: in update_source
    update_statements(self, self._old_statements, self._statements, trans)
../../src/pharmpy/plugins/nonmem/update.py:437: in update_statements
    warnings.warn(
E   UserWarning: Compartment structure has been updated, CMT-column in dataset might not be relevant anymore. Check CMT-column or drop column

``` 
I checked the start_model's datainfo, 'CMT' column is dropped, but in here https://github.com/pharmpy/pharmpy/blob/729a8f3e4dd76d6273a17181fa70089b953296c3/src/pharmpy/plugins/nonmem/update.py#L435  when I print col_dropped, it is not correct:
`{'ID': False, 'VISI': False, 'XAT2': False, 'DGRP': False, 'DOSE': False, 'FLAG': False, 'ONO': False, 'XIME': False, 'NEUY': False, 'SCR': False, 'AGE': False, 'SEX': False, 'NYHA': False, 'WT': False, 'COMP': False, 'ACE': False, 'DIG': False, 'DIU': False, 'NUMB': False, 'TAD': False, 'TIME': False, 'VIDD': False, 'CLCR': False, 'AMT': False, 'SS': False, 'II': False, 'VID': False, 'CMT': False, 'CONO': False, 'DV': False, 'EVID': False, 'OVID': False}`
I am not sure how to solve this problem, I think for the models which have datainfo file, maybe we don't need to use this function https://github.com/pharmpy/pharmpy/blob/729a8f3e4dd76d6273a17181fa70089b953296c3/src/pharmpy/plugins/nonmem/update.py#L434  to get drop information? 

Maybe we can have a discussion tomorrow?

Best regards,
Zhe Huang

ZheHuang96:
Thanks to Aurelien @pharmpy-dev-123 , we figure out the problem is caused by the drop column conflicts between datainfo file and $input record, e.g. in the $input record of mox2.mod, EVID=DROP but I didn't drop EVID in datainfo. A possible solution could be in Aurelien's latest PR: https://github.com/pharmpy/pharmpy/pull/892.
I think maybe for other cases we should prioritize $input record, but in amd tool, we start with a dataset and a datainfo,  we should use the datainfo file to drop columns?

rikardn:
Interesting. So the situation is that we have a conflict between what is stated in `.datainfo` and the NONMEM model file. I agree that for AMD there is really no conflict and we should always use the `.datainfo` since we do not (yet) have a model file. But for cases were we have a conflict we need to decide on what to do. Some options:

1. Always let .datainfo override the model.
    * Should then the parsing of the NONMEM dataset be done with the datainfo DROP?
    * We would need to update the $INPUT directly when reading
2. Always let the model override the datainfo. This feels strange
3. Raise saying that the model and datainfo does not match

To me 1. feels like the most natural, but should we then override all DROP that are in the model and only use the datainfo?

ZheHuang96:
I think there are three combinations in the input side:
1. **mod file + dataset**: I guess this will be the most common input combination, in this case, we use the DROP information from mod file.
2. **dataset + datainfo**: in this case, we should use the DROP information from datainfo file.
3. **mod file + dataset +datainfo** : If the mod file and datainfo have conflicts, it depends on which file users trust most? The mod file or the datainfo file? I guess most of the users are familiar with mod file rather than datainfo, if we use datainfo directly to handle DROP and update $INPUT of mod file, users could be panic? Maybe we should warn that the DROP information in datainfo and in the mod file don't match, and show the difference between two files at the first step? Then we can say we are going to let datainfo override the model, if they don't want to do that, they can stop the procedure and go back to check the files manually.

A workaround could be: We always let datainfo override the model, but we set up a pre-check function before it (maybe inside the `read_model` function?), when they have mod file and datainfo file, we check the DROP information between two files, so that user can do the check and get the information before doing any other things in pharmpy.

ZheHuang96:
Yes, I agree with you, I think **b** is our current 'best' option.

ZheHuang96:
> I thing 1 and 2 are clear.
> 
> For number 3 I can also see the problem with the user expectations, but I think we have to do something that is general for all Pharmpy tools. Ideally the datainfo and the model should not contradict and we need to make sure that whatever Pharmpy creates should not contain any contradictions. This doesn't stop external factors to create contradictions in the datainfo and in the model files though. **a)** Could we let the datainfo have priority and warn if there is a contradiction? **b)** We could also let the model have priority and update the datainfo accordingly with a warning. This option would potentially be less confusing to the users.
> 
> Option b would let users update everything that the model file supports in that file and everything else in the datainfo.
> 
> **c)** Raise
> 
> This option feels tough to the user. It means that dropping a column would have to be done both in the datainfo and in the modelfile if the user wants to do it manually. I am currently leaning towards option b. Any arguments against it (or for the other options)?


I think we still need to deal with the datainfo file problem before adding the amd test. Because of the time limitation, I think maybe I can copy and rename the csv file and also rename the datainfo file name so that we can deal with the conflicts later and only focus on solving this problem to set up amd test: **2**. dataset + datainfo: in this case, use the DROP information from datainfo file.
Or maybe we should entirely solve this problem since we have some solutions?

ZheHuang96:
Hi Rikard,

I think this PR should include everything we need for amd tool now. But maybe you can have a double check to make sure:) 
I used the smallest search space for amd test but it still has quite a long test time. Hopefully, the checks can pass so that you can start reviewing this PR.

Best regards,
Zhe Huang

rikardn:
Still some datainfo/model file mismatch error in the tests?

ZheHuang96:
I am sorry, it seems to have more columns than I suppose...I will make them match this weekend.

ZheHuang96:
Finally, it can pass our checks!

src/pharmpy/tools/run_amd.py
```diff
@@ -1,6 +1,7 @@
 from functools import partial
 
 import pandas as pd
+from sympy import Symbol
```
pharmpy-dev-123, line 4: Isn't the convention to use `sympy.Symbol` @rikardn?

src/pharmpy/tools/run_amd.py
```diff
@@ -216,25 +208,30 @@ def _run_resmod(model, path):
...
+        for col in model.datainfo:
+            if col.type == 'covariate' and col.continuous is False:
+                categorical.append(col.name)
+    cat_covariates = [Symbol(item) for item in categorical]
+
+    if continuous is not None or categorical is not None:
+        covariates_search_space = (
+            f'CONTINUOUS({con_covariates}); CATEGORICAL({cat_covariates})\n'
+            f'COVARIATE(@IIV, @CONTINUOUS, exp, *)\n'
+            f'COVARIATE(@IIV, @CATEGORICAL, cat, *)'
```
pharmpy-dev-123, line 229: You can also refer to all PK parameters with `*` instead of `@IIV`. Not sure which one is better @rikardn?

pharmpy-dev-123 (commented):
Some comments for @rikardn.

src/pharmpy/tools/run_amd.py
ZheHuang96: Previously, SCM use this IIV strategy, maybe this is our starting point?

rikardn: Perhaps for now `*` could mean the same as `@IIV`, but if we find new ways of detecting individual parameters we can enhance the `*` option.

PR #892, opened 2022-06-15
[fix] Make NONMEM use datainfo drop information in `_read_dataset`

pharmpy-dev-123:
TODO:

  - [ ] make sure this does not break other things


pharmpy-dev-123:
This will be superseded by upcoming commits.

PR #913, opened 2022-06-20
[docs] Some proofreading before the big release!

pharmpy-dev-123:
> @rikardn Ready to MERGE.


rikardn:
Looks great! I always use `-k` in pytest. Is there a difference? Will merge anyway

pharmpy-dev-123:
> Looks great! I always use `-k` in pytest. Is there a difference? Will merge anyway

When you stack targets it looks more declarative: `pytest a.py::test_a b.py::test_b`

PR #973, opened 2022-07-13
[docs] Merge COVsearch DSL documentation into MFL documentation

pharmpy-dev-123:


docs/modelsearch.rst
```diff
@@ -392,6 +392,8 @@ MFL support the following types of options to feature descriptions:
 +---------------+------------------+-------------------------------------------------------+
 | array         | :code:`[FO, ZO]` | Multiple tokens or numbers                            |
 +---------------+------------------+-------------------------------------------------------+
+| symbol        | :code:`@IIV`     | A symbol with manual or automatic definition          |
++---------------+------------------+-------------------------------------------------------+
 
 Model features
```
stellabelin, line 398: I was thinking since the MFL now covers more than one tool (and the idea is to extend to IIV etc, right?), then we can have this documentation in Tools under the subheading Common features?

docs/modelsearch.rst
```diff
@@ -412,6 +414,41 @@ MFL support the following model features:
...
+.. _mfl_symbols:
+
+Symbols
+~~~~~~~
+
+The MFL supports defining symbols via the syntax `LET(SYMBOL, [...])`.  For
+instance, to declare a list of absorption parameters use `LET(ABSORPTION, KA)`.
+Those symbols can then be referred to when declaring covariate effects via the
+syntax `COVARIATE(@SYMBOL, ...)`. Certain symbols have an automatic definition
+when not manually defined:
```
stellabelin, line 431: To me this paragraph is a little confusing. If I understand it correctly, you use the LET syntax to say that when we refer to ABSORPTION, KA is the parameter we mean. I was thinking that the order of the paragraph might make it a little confusing for me. The order that feels more logical for me would be first to explain what a symbol is, then write that Pharmpy automatically "decides" the symbols, then the table, and then write that you can also manually declare the different types of parameters.

docs/modelsearch.rst
```diff
@@ -502,3 +539,10 @@ Allow all combinations of absorption and elimination rates:
 
     ABSORPTION(*)
     ELIMINATION(*)
+
+All covariate effects on parameters with IIV:
+
+.. code::
+
+    COVARIATE(@IIV, @CONTINUOUS, *)
+    COVARIATE(@IIV, @CATEGORICAL, CAT)
```
stellabelin, line 548: Maybe an example of defining symbols? Something I was also thinking of is that it could be could to have a list example since lists have different syntax is Python and in R.

docs/modelsearch.rst
```diff
@@ -412,6 +414,41 @@ MFL support the following model features:
...
+| `@ABSORPTION`   | Parameter   | auto or manual | All PK absorption parameters                |
++-----------------+-------------+----------------+---------------------------------------------+
+| `@ELIMINATION`  | Parameter   | auto or manual | All PK elimination parameters               |
++-----------------+-------------+----------------+---------------------------------------------+
+| `@DISTRIBUTION` | Parameter   | auto or manual | All PK distribution parameters              |
++-----------------+-------------+----------------+---------------------------------------------+
+| `@CONTINUOUS`   | Covariate   | auto or manual | All continuous covariates                   |
++-----------------+-------------+----------------+---------------------------------------------+
+| `@CATEGORICAL`  | Covariate   | auto or manual | All categorical covariates                  |
++-----------------+-------------+----------------+---------------------------------------------+
```
stellabelin, line 447: Since all of the aliases can be either automatically or manually defined, would it be clearer without the Definition column?

stellabelin (commented):
Looks good! I've added some comments

docs/modelsearch.rst
pharmpy-dev-123: Yep that makes sense. I will move it.

docs/modelsearch.rst
pharmpy-dev-123: Yep. And I'll rename Alias to Symbol in the header.

docs/modelsearch.rst
pharmpy-dev-123: OK. I will rewrite.

PR #1012, opened 2022-07-29
Amd change

ZheHuang96:


tests/integration/test_amd.py
```diff
@@ -20,5 +20,4 @@ def test_amd(tmp_path, testdata):
         )
         assert len(res.summary_tool) == 58
         assert len(res.summary_models) == 60
-        assert len(res.summary_individuals_count) == 60
```
pharmpy-dev-123, line 23: Why is this removed?

src/pharmpy/tools/run_amd.py
```diff
@@ -126,48 +126,64 @@ def run_amd(
...
                 _run_covariates, continuous=continuous, categorical=categorical, path=db.path
             )
-            run_subfuncs['covsearch'] = func
+            subresults = func(next_model)
+            next_model = subresults.best_model
+            run_subtools['covsearch'] = subresults
         else:
             raise ValueError(
                 f"Unrecognized section {section} in order. Must be one of {default_order}"
             )
```
pharmpy-dev-123, line 168: This will not allow to store subresults if a tool is present multiple times in `order`. Should use a list instead of a dictionary?

pharmpy-dev-123 (changes requested):

pharmpy-dev-123 (approved):

PR #1015, opened 2022-08-01
Fix rate column issue (fixes #989)

### LINKED ISSUE #989: Fail to read in dataset when there is RATE column in the dataset
When the input is a dataset, Pharmpy cannot recognize the column header "RATE". The error information is:
image
However, RATE is also a special variable name in NONMEM. If the RATE is changed to other strings like RT or RTE, then NONMEM cannot recognize it. The RATE plays an important role in calculating the PK of iv infusion drugs, the parameter estimation result will be largely affected if the RATE data cannot not used.

ZheHuang96:
Hi both,

@pharmpy-dev-123 and I found the issue was made by `nonmem.plugin.model.convert_model()`. Because we need to use the dataset to change the model statements (dose type: bolus to infusion), but the path is not correct: `file.csv`. So we need to set the correct path before we change the statement. Then another issue pops out:
```
Traceback (most recent call last):

  Input In [5] in <cell line: 4>
    run_amd('C:\\Users\\zhehu311\\Desktop\\debug\\amd\\xuanlin\\dauno_filtered.csv',

  File ~\Anaconda3\lib\site-packages\pharmpy\tools\run_amd.py:102 in run_amd
    model = convert_model(model, 'nonmem')  # FIXME: Workaround for results retrieval system

  File ~\Anaconda3\lib\site-packages\pharmpy\modeling\common.py:238 in convert_model
    new = nonmem.convert_model(model)

  File ~\Anaconda3\lib\site-packages\pharmpy\plugins\nonmem\model.py:96 in convert_model
    nm_model.update_source()

  File ~\Anaconda3\lib\site-packages\pharmpy\plugins\nonmem\model.py:244 in update_source
    datapath = write_csv(self, path=dir_path, force=force)

  File ~\Anaconda3\lib\site-packages\pharmpy\modeling\write_csv.py:42 in write_csv
    raise FileExistsError(f'File at {path} already exists.')

FileExistsError: File at start.csv already exists.
```
I use the argument: `nofiles=True` in `update_source()` to solve this issue. But I just checked the start.csv dataset from the modelfit directory, the RATE column disappear...

Please have a review when you have time and if you have any ideas, please tell me!

### DIFF
--- a/src/pharmpy/plugins/nonmem/model.py
+++ b/src/pharmpy/plugins/nonmem/model.py
@@ -74,6 +74,7 @@ def convert_model(model):
         code += '$PK\nY=X\n'
         code += '$ERROR\nA=B'
     nm_model = pharmpy.model.Model.create_model(StringIO(code))
+    nm_model._datainfo = model.datainfo
     nm_model.random_variables = model.random_variables
     nm_model._parameters = model.parameters
     nm_model._old_parameters = Parameters()
@@ -85,7 +86,6 @@ def convert_model(model):
     nm_model.value_type = model.value_type
     nm_model._data_frame = model.dataset
     nm_model._estimation_steps = model.estimation_steps
-    nm_model._datainfo = model.datainfo
     nm_model._initial_individual_estimates = model.initial_individual_estimates
     nm_model.observation_transformation = model.observation_transformation.subs(
         model.dependent_variable, nm_model.dependent_variable
--- a/src/pharmpy/plugins/nonmem/advan.py
+++ b/src/pharmpy/plugins/nonmem/advan.py
@@ -451,10 +451,14 @@ def _advan12_trans(trans):
 
 
 def _dosing(model, dose_comp):
-    di = model.datainfo
+    return dosing(model.datainfo, lambda: model.dataset, dose_comp)
+
+
+def dosing(di, dataset, dose_comp):
     colnames = di.names
+
     if 'RATE' in colnames and not di['RATE'].drop:
-        df = model.dataset
+        df = dataset()
         if (df['RATE'] == 0).all():
             return Bolus('AMT')
         elif (df['RATE'] == -1).any():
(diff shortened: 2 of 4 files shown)


stellabelin (commented):
Hi! So it enters the following if-statement? https://github.com/pharmpy/pharmpy/blob/362f7a2e90c5e82546a7719ecad68bd94875d3e5/src/pharmpy/plugins/nonmem/model.py#L236 Shouldn't `self.datainfo.path` be the `datainfo.path` of `model`, i.e. the one from the input datainfo-file?

ZheHuang96:
> Hi! So it enters the following if-statement?
> 
> https://github.com/pharmpy/pharmpy/blob/362f7a2e90c5e82546a7719ecad68bd94875d3e5/src/pharmpy/plugins/nonmem/model.py#L236
> 
> Shouldn't `self.datainfo.path` be the `datainfo.path` of `model`, i.e. the one from the input datainfo-file?

Yes...I think we figure out the reason, because the dataset has been changed (drop the RATE column), so the datainfo has been reset. @pharmpy-dev-123 has an idea to fix this issue and we will give a PR later.

src/pharmpy/plugins/nonmem/model.py
```diff
@@ -85,13 +86,12 @@ def convert_model(model):
...
     nm_model._data_frame = model.dataset
     nm_model._estimation_steps = model.estimation_steps
-    nm_model._datainfo = model.datainfo
     nm_model._initial_individual_estimates = model.initial_individual_estimates
     nm_model.observation_transformation = model.observation_transformation.subs(
         model.dependent_variable, nm_model.dependent_variable
     )
     nm_model.description = model.description
-    nm_model.update_source()
+    nm_model.update_source(nofiles=True)
```
pharmpy-dev-123, line 94: This skips an unsuccessful file overwrite attempt that occurs when the dataset is updated (for instance when we remove the RATE column from the dataframe). This is not needed if we make sure the RATE column does not get deleted.

PR #1029, opened 2022-08-08
[perf] Speed-up unit tests execution

### LINKED ISSUE #1136: Speed up subs method
@pharmpy-dev-123 found that 5% of time in unit tests was spent in `statements.subs` so we should try to optimize.

Some ideas:

1. Add an `xreplace` method and use that in applicable cases (most?).

2. Create an even faster method that only replace symbols

3. Use symengine for statements. There are some incompatibilities between sympy and symengine that need to be worked around. Either by confining the use of symengine to certain classes and convert to sympy when going outside, by some wrapping functions or by fixing the issues directly in symengine.py

pharmpy-dev-123:
> @rikardn REVIEW. I do not see how to fix the leftover bullet points for `conditional_joint_normal`.

### Speed-up individual parameter statistics calculations

Current speed-up on my machine is roughly x15 (120s at 194e544, 8s with this PR)

```sh
tox -e py39 -- -n0 -qq tests/modeling/test_results.py::{test_calc_pk_two_comp_bolus,test_calculate_pk_parameters_statistics,test_calculate_individual_parameter_statistics}
```

Tricks used:
  - [x] 1. Reuse expensive `sympy` `MultivariateNormal` objects when possible (by caching the output of `RandomVariable#sympy_rv` but also through manual management).
  - [x] 2. Use `xreplace` instead of `subs` (see https://github.com/sympy/sympy/issues/22240)
  - [x] 3. Avoid creating many different lambdified expressions by passing the parameter samples as parameters to the lambda instead of incorporating them into the lambdification. Two-stage sampling is unavoidable because we first sample the OMEGAs, then we sample the ETAs, so it also makes abstracting more difficult.
  - [x] 4. Sample directly through `numpy`. Makes point 1 kind almost moot. We could maybe refactor back to higher-level abstraction `RandomVariables#sample`. Problem is how to make point 3 work with a higher-level abstraction. And we could also we want to refactor to a lower-level abstraction to avoid the use of `functools.lru_cache` that cache expensive `lambdify` calls.
  - [x] 5. First convert sympy matrices to symengine, then call `xreplace`.
  - [x] 6. Hard-code individual symbol replacements by dict lookup.
  - [x] 7. Reuse samples when evaluating multiple expressions simultaneously.
  - [x] 8. Project multivariate normals down to what is strictly necessary for the current sampling.

### Speed-up `subs`

Brought down overall test usage of `subs` from > 10% to ~2%.

  - [x] Fix #1136
  - [x] Shortcut when substitution dictionary is "simple".

### Speed-up `conditional_joint_normal`

  - [x] ~Reuse intermediate computation.~ Merged via #1156
  - [ ] Find a way to reuse intermediate values in loops using `conditional_joint_normal`
  - [ ] Find a way to get rid of matrix inversion

### Speed-up `RandomVariable.normal`

  - [x] ~Directly query variance for nonnegativity, instead of asking for positive-semidefiniteness of 1x1 covariance matrix.~ Merged via #1156

### Speed-up `get_unit_of`

  - [x] Filter unit equations before passing them to `sympy.solve`.


src/pharmpy/random_variables.py
```diff
@@ -1185,7 +1185,10 @@ def _sample_expr_from_rvs(sampling_rvs, expr, parameters, samples, rng):
 
     ordered_symbols, fn = _lambdify_canonical(expr)
     data = _sample_rvs_subset(sampling_rvs, ordered_symbols, samples, rng)
-    return fn(*data)
+    results = fn(*data)
+    # FIXME Why is this necessary? Was not before sympy 1.11.0 and is not with
+    # cse=False.
+    return results[0] if isinstance(results, list) else results
```
pharmpy-dev-123, line 1213: See https://github.com/sympy/sympy/pull/23538#issuecomment-1225962565

rikardn:
I think we should wait with this until the redesign of `RandomVariables` since this PR changes the top level API it should instead use the new API. I think you are right that a higher focus on the distributions rather than the random variables is warranted since there is a tight coupling within a distribution. So `RandomVariables` should contain `Distribution`s

src/pharmpy/random_variables.py
pharmpy-dev-123: https://github.com/sympy/sympy/pull/23975 was merged and [the patch is included in sympy 1.11.1](https://github.com/sympy/sympy/wiki/release-notes-for-1.11.1#changes). I'll rebase, edit, and resolve later.

pharmpy-dev-123: Reverted since the bug was fixed upstream.

rikardn (approved):
Good catches!

rikardn:
Sorry. Only reviewed the last commit. Perhaps this and other commits could be split out so we could merge them separately?

src/pharmpy/expressions.py
```diff
@@ -56,3 +56,142 @@ def _free_images_iter(expr: sympy.Expr) -> Iterable[sympy.Expr]:
...
+    )
+
+
+@lru_cache(maxsize=256)
+def _sympify_new(new) -> sympy.Expr:
+    # NOTE This mimics sympy's input coercion in subs
+    return sympy.sympify(new, strict=not isinstance(new, (str, type)))
+
+
+def subs(expr: sympy.Expr, mapping: Dict[sympy.Expr, sympy.Expr], simultaneous: bool = False):
```
rikardn, line 80: Do we need the `simultaneous` option? In my mind it should always be `True`, but perhaps there are use cases for `False` that I couldn't think of.

src/pharmpy/expressions.py
```diff
@@ -56,3 +56,142 @@ def _free_images_iter(expr: sympy.Expr) -> Iterable[sympy.Expr]:
...
+        return sympy.Integer(0)
+    if fn is sympy.Mul:
+        return sympy.Integer(1)
+    if fn is sympy.Pow:
+        return sympy.Integer(1)
+
+    raise ValueError(f'{type(fn)}: {repr(fn)}')
+
+
+def prune(predicate: Callable[[sympy.Expr], bool], expr: sympy.Expr):
```
rikardn, line 166: What does this function do? Perhaps a small comment to explain.

src/pharmpy/model/random_variables.py
```diff
@@ -137,15 +148,35 @@ def subs(self, d):
...
+        mean = subs(self._mean, d)
+        variance = subs(self._variance, d)
         name = self._names[0]
         if name in d or sympy.Symbol(name) in d:
             name = d.get(name, d.get(sympy.Symbol(name)))
             if isinstance(name, sympy.Symbol):
                 name = name.name
         return NormalDistribution((name,), self._level, mean, variance)
 
+    def __getitem__(self, index):
```
rikardn, line 160: Interesting idea. What will be returned by indexing?

src/pharmpy/distributions.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
+
+from abc import abstractmethod
+from typing import Union
+
+from pharmpy.deps import numpy as np
+
+
+class NumericDistribution:
```
rikardn, line 9: Do we still need these? They are very similar to the "symbolic" distribution classes.

pharmpy-dev-123: What I use is the polymorphism of the `sample` method. It makes sense to build intermediate objects since substituting symbols is so costly (I have not solved that problem yet because it is the `.subs` method of `sympy` `Matrix` objects, and also it's better to reuse built `numpy` objects).

pharmpy-dev-123: The names should be improved.

src/pharmpy/model/random_variables.py
pharmpy-dev-123: The marginal!

src/pharmpy/expressions.py
pharmpy-dev-123: It removes nodes from expressions. I use it instead of `expr.subs({sympy.exp: 1})` which replaces `e^x` subexpressions by `1`. I'll add a docstring.

src/pharmpy/expressions.py
pharmpy-dev-123: I wanted this to be a drop-in replacement for `.subs`. Some current uses of `subs` seemed to need `simultaneous = False` but this can be revisited.

The advantage of having a `subs` function that can do everything correctly and some things faster than `.subs` is that it can be used everywhere without thinking. If we manage to have a faster `subs` even when `simultaneous` is `False` then we won't have to change anything to benefit from the speed-up.

pharmpy-dev-123: Another example where a fast non-simultaneous `subs` could be useful is when computing `full_expression`.

src/pharmpy/expressions.py
pharmpy-dev-123: > I'll add a docstring.

Done.

src/pharmpy/distributions.py
pharmpy-dev-123: I gave a try at a new polymorphic API to create numeric distribution from symbolic ones: `numeric_dist = symbolic_dist.evalf(parameters)` and then `numeric_dist.sample(...)`.

PR #1035, opened 2022-08-11
[dev] Speed up tests by caching model parsing

pharmpy-dev-123:
> Ready to MERGE @rikardn 

This PR speeds up model parsing in tests by caching the parsed model and
generate copies with `Model#copy` instead. Tests run significantly faster but
faster execution could still be achieved by improving the parsing speed: many
tests parse small pieces of NMTRAN code that are used nowhere else, so they
cannot be cached. So I have not made these tests use the caching mechanism.

Some caveats:
  - I had to make `Model#copy` work with models that do not have a name.
  - `Model.create_model`'s output depends on Pharmpy's configuration, so the
cache key is currently made to depend on the string representation of the
NONMEM part of the config. This could maybe made more generic. I also noticed
that the default config is represented by the empty string, but as soon as you
use a config context the representation changes to an explicit stringification
of the defaults (i.e. not the empty string).
  - Caching currently is path-based, not content-based, so an attempt is made
at resolving the given file path, and if that file path is a descendant of
pytest's basetemp directory, caching is skipped.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+from typing import Dict, Hashable, Optional, Union
 
 import pytest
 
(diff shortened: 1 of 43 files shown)


tests/conftest.py
```diff
@@ -20,7 +21,88 @@ def pheno_path(datadir):
...
-def pheno(pheno_path):
-    import pharmpy
+def pheno(load_model, pheno_path):
+    return load_model(pheno_path)
 
-    return pharmpy.Model.create_model(pheno_path)
+
+@pytest.fixture(scope='session')
+def load_model(tmp_path_factory):
+
```
stellabelin, line 30: Since `load_model` has such a similar name to `read_model` and `load_example_model`, I'm thinking that it might make it a little confusing since it sounds like it's a regular `modeling`-function. Perhaps it would be clearer to have it named something like: `load_cached_model` or `load_fixture_model`?

stellabelin: Same goes for `load_example` and `create_model`

stellabelin (commented):
Some small comments, I was also wondering: how much faster was it with fixtures with copy?

tests/conftest.py
pharmpy-dev-123: > I was also wondering: how much faster was it with fixtures with copy?

I do not understand?!

pharmpy-dev-123: Yes, I will think about better names!

stellabelin: I meant how much did this speed up?

pharmpy-dev-123: On my machine:
  - `tox -e py39 -- pytest -n4 --asyncio-mode=auto --cov --cov-report=term-missing --cov-report xml:.coverage.xml -vv tests`
    - 9m21s  at 43dcfa94 
    - 8m15s with #1035.
  - `tox -e py39 -- pytest -n8 tests/`
    - 4m47  at 43dcfa94 
    - 4m24 with #1035.
  
On GHA the tests consume a total of ~3.5 hours instead of ~4.25 hours.

pharmpy-dev-123: Names have been changed:
  - `load_model` to `load_model_for_test`
  - `load_example` to `load_example_model_for_test`
  - `create_model` to `create_model_for_test_with_pheno_dataset`

tests/conftest.py
```diff
@@ -89,7 +89,7 @@ def _parse_model():
 
 
 @pytest.fixture(scope='session')
-def create_model(load_example):
+def create_model_for_test_with_pheno_dataset(load_example_model_for_test):
```
ZheHuang96, line 92: It looks really nice to me and I have no comment but a random thought, maybe create_model_for_test is enough? It sounds more general and maybe we will add more test datasets in the future?

pharmpy-dev-123: Done!

ZheHuang96 (commented):
It looks very nice for me!

PR #1050, opened 2022-08-16
Add IIV_MAT as default when creating oral type start model

ZheHuang96:
Hi Stella,

I tried to add IIV on MAT based on our current pharmpy code many times and I got the same results. The previous failure is here: https://github.com/pharmpy/pharmpy/runs/7692944105?check_suite_focus=true Different systems can generate different numbers of models which is really weird. I think maybe it is also related to IIV candidate model names are not consistent, which we have already solved. If it is, then it is easy for us:). I think we need some proof, do we have enough quota to run the integration test via github? Is there an alternative way to test it between different systems? 

Best regards,
Zhe Huang


stellabelin (commented):
Hi! Yeah, the sorted order of IIV-candidates seems to have fixed the sporadic error at least. I can't find how to run the integration tests on this PR (can see for dependabot's PRs), @rikardn: do you know?

rikardn:
I think either locally by pulling this branch or on @ZheHuang96's fork.

ZheHuang96:
Thank you for your reply! I am trying to run the integration tests on my fork:)

ZheHuang96:
https://github.com/ZheHuang96/pharmpy/runs/7978296990?check_suite_focus=true#step:4:212
Unfortunately, the integration test can not be run on my own fork because of the wrong password of nonmem.

stellabelin:
Since the number of models can change quite a lot, maybe we shouldn't test the number of models? Or just test whether or not it is equal or more than a number, or within a range? We could also test that each subtool has created at least one model. Currently, if something is changed it can have a cascading effect on the number of models created, e.g. if an early tool chooses a more complex model with several IIVs, the subsequent search space for e.g. IIVSearch will be much larger, which leads to the IOVSearch having a bigger search space etc. Also with the new ranking system, different OS can get different errors and thus excluding models differently.

ZheHuang96:
Hi @stellabelin, thank you for your reply! We definitely need to find a good way to test the amd function.  And yes, I agree with you. Maybe I can start to test 'At least one model generated per subtool' to see what happens. By the way, may I know what is our new ranking system?

stellabelin:
I think that makes sense as a first test! The new system will be excluding models that have not minimized (except for rounding errors) from the ranking so that we only propagate models that minimized! It's the following issues: #1056 and #1055

ZheHuang96:
Nice, thank you for your reply. I will change the test today.

ZheHuang96:
Hi @stellabelin, 

The latest commit [6ab50be]  is the new test of amd, it follows the same idea as modelsearch test. Please have a look when you have time:)

stellabelin (approved):
Looks good!

ZheHuang96:
Ok, I will merge it.

PR #1066, opened 2022-08-21
[fix/ux] Use different port for dask distributed dashboard

### LINKED ISSUE #840: Silence port userwarning
The warning `UserWarning: Port 8787 is already in use` should be silenced in R since Rstudio uses 8787 and dask will use another if 8787 is taken, or dask should use another port.

pharmpy-dev-123:
> Ready to MERGE @rikardn 

The default port (8787) causes user warnings because it clashes with one of the ports used by RStudio. Fixes #840.

### DIFF
--- a/src/pharmpy/workflows/dispatchers/local_dask.py
+++ b/src/pharmpy/workflows/dispatchers/local_dask.py
@@ -33,7 +33,7 @@ def run(workflow):
                     # Catch deprecation warning from python 3.10 via tornado.
                     # Should be fixed with tornado 6.2
                     warnings.filterwarnings("ignore", message="There is no current event loop")
-                    with LocalCluster(processes=False) as cluster:
+                    with LocalCluster(processes=False, dashboard_address=':31058') as cluster:
                         with Client(cluster) as client:
                             print(client)
                             dsk_optimized = optimize_task_graph_for_dask_distributed(client, dsk)


stellabelin (commented):
Okay, looks good! I guess we need to look into other alternatives if we want to use the dashboard?

pharmpy-dev-123:
> I guess we need to look into other alternatives if we want to use the dashboard?

Yep. I would wait until we use it.

stellabelin:
Okay, makes sense!

rikardn:
We use it already now. Or rather we can use it and have used it. Is the warning serious or can it simply be silenced?

stellabelin:
The warning itself is not serious, in Rstudio it will always warn since Rstudio by default uses the default port for dask, it will warn and write which port it is using instead

rikardn:
Ok, so this is communication to the user of the dashboard to know which port to connect to. Could we use another port that is less likely to clash with Rstudio?

pharmpy-dev-123:
If it is used by the modeler then we need to leave the warning in place so that they know the port is not the expected one. And to improve the UX for modelers using RStudio we should choose a different port than 8787 so that the warning happens less often.

pharmpy-dev-123:
The full warning message looks like the following:

```
.../site-packages/distributed/node.py:160: UserWarning: Port 8787 is already in use.
Perhaps you already have a cluster running?
Hosting the HTTP server on port 54109 instead
```

So it gives the used port when not using the expected one and we do not need to do anything except select and advertise a different port.

pharmpy-dev-123:
I edited this PR to use a different default port instead of disabling the service.

PR #1067, opened 2022-08-21
[fix] Correctly handle parameter discovery in `add_allometry`

### LINKED ISSUE #1044: run_allometry error: ValueError('No parameters found')
This error occurred when testing the `run_amd` with warfarin dataset. Below is the detailed error information:
```
Key:       add allometry-run0-results
Function:  execute_task
args:      ((<function results at 0x2abccf7f6160>, <Pharmpy model object best_resmod_1>, (<function retrieve_from_database_or_execute_model_with_tool.<locals>.task at 0x2abd31d38a60>, (functools.partial(<function _add_allometry_on_model at 0x2abccf7f60d0>, allometric_variable='WT', reference_value=70, parameters=None, initials=None, lower_bounds=None, upper_bounds=None, fixed=True), <Pharmpy model object best_resmod_1>))))
kwargs:    {}
Exception: "ValueError('No parameters found')"
```
I then checked the AMD user guide. The description in the "AMD" (https://pharmpy.github.io/latest/amd.html#allometry) is different from those specified in the "allometry" (https://pharmpy.github.io/latest/allometry.html#running). The default setting of parameters to apply scaling to is supposed be all CL, Q and V parameters but it is `None` in this case, so do `initials`, `lower_bounds`, and `upper_bounds`.

There are no detailed arguments in the example code but I tried to write by myself and tested the `run_allometry` subtool individually, using the final model from `resmod` module as an input. Below is the input model and my code:

Input model
```
$PROBLEM [VC,CL]+[MAT,MDT]
$INPUT ID TIME WT AGE SEX AMT DVID DV MDV
$DATA best_resmod_1.csv IGNORE=@  
$SUBROUTINES ADVAN5  TRANS1 
$MODEL COMPARTMENT=(TRANSIT1 DEFDOSE) COMPARTMENT=(TRANSIT2) COMPARTMENT=(TRANSIT3) COMPARTMENT=(TRANSIT4) COMPARTMENT=(TRANSIT5) COMPARTMENT=(TRANSIT6) COMPARTMENT=(TRANSIT7) COMPARTMENT=(TRANSIT8) COMPARTMENT=(TRANSIT9) COMPARTMENT=(TRANSIT10) COMPARTMENT=(DEPOT) COMPARTMENT=(CENTRAL)
$PK
MDT = THETA(4)*EXP(ETA(4))
VC = THETA(1) * EXP(ETA(1))
CL = THETA(2) * EXP(ETA(2))
MAT = THETA(3) * EXP(ETA(3))
KA = 1/MAT
V = VC
K12 = 10/MDT
K23 = 10/MDT
K34 = 10/MDT
K45 = 10/MDT
K56 = 10/MDT
K67 = 10/MDT
K78 = 10/MDT
K89 = 10/MDT
K9T10 = 10/MDT
K10T11 = 10/MDT
K11T12 = KA
K12T0 = CL/V
$ERROR
IF (A(12)/VC.EQ.0) THEN
    IPREDADJ = 2.22500000000000E-16
ELSE
    IPREDADJ = A(12)/VC
END IF
Y = A(12)/VC + A(12)*EPS(1)/VC + EPS(2)
$ESTIMATION METHOD=COND INTERACTION MAXEVAL=9999 SADDLE_RESET=1 
$TABLE ID TIME CWRES CIPREDI VC DV PRED RES WRES IPREDI IWRESI FILE=mytab NOAPPEND NOPRINT
$THETA (0, 8.01621, Inf) ; POP_VC
$THETA (0, 0.131785, Inf) ; POP_CL
$THETA (0, 0.445779, Inf) ; POP_MAT
$THETA  (0,0.987272) ; POP_MDT
$OMEGA BLOCK(2)
0.0479822	; IIV_VC
0.0258434	; IIV_VC_IIV_CL
0.0783349	; IIV_CL
$OMEGA BLOCK(2)
1.89185	; IIV_MAT
-0.676248	; IIV_MAT_IIV_MDT
0.391653	; IIV_MDT
$SIGMA  0.702575 ; sigma_prop
$SIGMA  2.49449 ; sigma_add
```
Testing code
```
library(pharmr)

model <- read_model('best_resmod_1.mod')
fit(model)

res <- run_allometry(model=model,
                     allometric_variable = 'WT',
                     reference_value = 70,
                     parameters = list('VC', 'CL'),
                     initials = c(1, 0.75))
```
And this code works. So I suppose this error is related to the setting of `parameters` and `initials`. 

I don't know if it is possible to specify these two arguments in the `run_amd` function. Also, I am not sure why this only happened in one single dataset, other datasets also accept the same default setting, which is `None`, but do not have this error.

### LINKED ISSUE #1053: Support finding clearance and volume parameters when there is a transit compartment or MM elimination
transit compartment issue: https://github.com/pharmpy/pharmpy/issues/1044
MM issue: https://github.com/pharmpy/pharmr/issues/89

Some functions may need to improve:
`find_clearance_parameters` ,`find_volume_parameters` or maybe `get_flow`

### LINKED ISSUE #1062: Issues with allometry tool and Questions
Issue 1.
I used run_allometry on a best model from a previous modelsearch that is a 3 CPT model with MM elimination. I get an error message:

distributed.worker - WARNING - Compute Failed
 Function:  _add_allometry_on_model
 args:      (<Pharmpy model object modelsearch_candidate10>)
 kwargs:    {}
 Exception: StopIteration()

Error in py_call_impl(callable, dots$args, dots$keywords) :
    RuntimeError: coroutine raised StopIteration

I am guessing the tool does not recognize CLMM as a parameter. For me it should be possible to add allometry on CLMM as on CL

Issue 2.

I do not know how to limit or define the parameters on which adding allometric scaling, I have tried:
 parameters = c("CL") or c(CL) or list("CL") or "CL" or CL, nothing works and get error message.

Questions

Can we have the same exponent for all CLs and same exponent for all Vs (but different from the CLs)?
Can the reference_value be automatically set to the median of allometric_varaible in the dataset?

pharmpy-dev-123:
> @rikardn Ready to MERGE

The goal is to fix the following issues:
  - [x] Handle complex rates in `add_allometry`
  - [x] Fixes #1044
  - [x] Fixes #1062
  - [x] Fixes #1053

### DIFF
--- a/src/pharmpy/statements.py
+++ b/src/pharmpy/statements.py
@@ -1255,7 +1255,7 @@ class CompartmentalSystem(ODESystem):
         derivatives = sympy.Matrix([sympy.Derivative(fn, self.t) for fn in amount_funcs])
         inputs = self.zero_order_inputs
         a = self.compartmental_matrix @ amount_funcs + inputs
-        eqs = [sympy.Eq(lhs, rhs) for lhs, rhs in zip(derivatives, a)]
+        eqs = [sympy.Eq(lhs, simplify_ode_rhs(rhs)) for lhs, rhs in zip(derivatives, a)]
         ics = {}
         output = self.output_compartment
         for node in self._g.nodes:
@@ -1346,6 +1346,25 @@ class CompartmentalSystem(ODESystem):
         dose = self.dosing_compartment.dose
         s = str(dose) + '\n' + str(grid).rstrip()
         return s
+
+
+def simplify_ode_rhs(expr: sympy.Expr):
+    fi = _free_images(expr)
+    return sympy.collect(_expand_rates(expr, fi), sorted(fi, key=str))
+
+
+def _expand_rates(expr: sympy.Expr, free_images: Set[sympy.Expr]):
+    if isinstance(expr, sympy.Add):
+        return sympy.expand(
+            sympy.Add(*map(lambda x: _expand_rates(x, free_images), expr.args)), deep=False
+        )
+    if (
+        isinstance(expr, sympy.Mul)
+        and len(expr.args) == 2
+        and not free_images.isdisjoint(expr.args)
+    ):
+        return sympy.expand(expr, deep=False)
+    return expr
 
 
 class Compartment:
--- a/src/pharmpy/statements.py
+++ b/src/pharmpy/statements.py
@@ -424,10 +424,8 @@ class ExplicitODESystem(ODESystem):
             cb.add_compartment(comp)
 
         for eq in self.odes:
-            for comp_func in funcs:
+            for comp_func in _free_images(eq.rhs):
                 dep = eq.rhs.as_independent(comp_func, as_Add=True)[1]
-                if dep == 0:
-                    continue
                 terms = sympy.Add.make_args(dep)
                 for term in terms:
                     if _is_positive(term):
--- a/src/pharmpy/statements.py
+++ b/src/pharmpy/statements.py
@@ -1,20 +1,19 @@
 import copy
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
-from functools import reduce
-from itertools import chain
-from operator import __and__
-from typing import Iterable, Set
 
 import networkx as nx
 import sympy
 
 import pharmpy.unicode as unicode
 
-
-def sympify(expr):
-    ns = {'Q': sympy.Symbol('Q')}
-    return sympy.sympify(expr, locals=ns)
+from .expressions import (
+    assume_all,
+    canonical_ode_rhs,
+    free_images,
+    free_images_and_symbols,
+    sympify,
+)
 
 
 class Statement(ABC):
@@ -422,7 +421,7 @@ class ExplicitODESystem(ODESystem):
             compartments[name] = comp
 
         for eq in self.odes:
-            for comp_func in concentrations.intersection(_free_images(eq.rhs)):
+            for comp_func in concentrations.intersection(free_images(eq.rhs)):
                 dep = eq.rhs.as_independent(comp_func, as_Add=True)[1]
                 terms = sympy.Add.make_args(dep)
                 for term in terms:
@@ -592,34 +591,8 @@ class CompartmentalSystemBuilder:
 
 def _is_positive(expr: sympy.Expr) -> bool:
     return sympy.ask(
-        sympy.Q.positive(expr), _assume_all(sympy.Q.positive, _free_images_and_symbols(expr))
+        sympy.Q.positive(expr), assume_all(sympy.Q.positive, free_images_and_symbols(expr))
     )
-
-
-def _free_images_and_symbols(expr: sympy.Expr) -> Set[sympy.Expr]:
-    return expr.free_symbols | _free_images(expr)
-
-
-def _free_images(expr: sympy.Expr) -> Set[sympy.Expr]:
-    return set(_free_images_iter(expr))
-
-
-def _free_images_iter(expr: sympy.Expr) -> Iterable[sympy.Expr]:
-    if isinstance(expr.func, sympy.core.function.UndefinedFunction):
-        yield expr
-        return
-
-    yield from chain.from_iterable(
-        map(
-            _free_images,
-            expr.args,
-        )
-    )
-
-
-def _assume_all(predicate: sympy.assumptions.Predicate, expressions: Iterable[sympy.Expr]):
-    tautology = sympy.Q.is_true(True)
-    return reduce(__and__, map(predicate, expressions), tautology)
 
 
 class CompartmentalSystem(ODESystem):
@@ -1249,7 +1222,7 @@ class CompartmentalSystem(ODESystem):
         derivatives = sympy.Matrix([sympy.Derivative(fn, self.t) for fn in amount_funcs])
         inputs = self.zero_order_inputs
         a = self.compartmental_matrix @ amount_funcs + inputs
-        eqs = [sympy.Eq(lhs, simplify_ode_rhs(rhs)) for lhs, rhs in zip(derivatives, a)]
+        eqs = [sympy.Eq(lhs, canonical_ode_rhs(rhs)) for lhs, rhs in zip(derivatives, a)]
         ics = {}
         output = self.output_compartment
         for node in self._g.nodes:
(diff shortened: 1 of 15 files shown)


src/pharmpy/statements.py
```diff
@@ -593,6 +590,38 @@ def set_lag_time(self, compartment, lag_time):
...
+def _free_images_and_symbols(expr: sympy.Expr) -> Set[sympy.Expr]:
+    return expr.free_symbols | _free_images(expr)
+
+
+def _free_images(expr: sympy.Expr) -> Set[sympy.Expr]:
+    return set(_free_images_iter(expr))
+
+
+def _free_images_iter(expr: sympy.Expr) -> Iterable[sympy.Expr]:
+    if isinstance(expr.func, sympy.core.function.UndefinedFunction):
```
rikardn, line 608: Is this needed because we can have symbolic functions in expressions?

src/pharmpy/statements.py
```diff
@@ -1310,6 +1342,25 @@ def __repr__(self):
         return s
 
 
+def simplify_ode_rhs(expr: sympy.Expr):
```
rikardn, line 1345: This function would become part of the model API. What does it do?

src/pharmpy/statements.py
```diff
@@ -593,6 +590,38 @@ def set_lag_time(self, compartment, lag_time):
         return new_comp
 
 
+def _is_positive(expr: sympy.Expr) -> bool:
```
rikardn, line 593: It assumes every symbol to be positive which might not be the case. What if there is a covariate in the system or a theta?

src/pharmpy/statements.py
```diff
@@ -593,6 +590,38 @@ def set_lag_time(self, compartment, lag_time):
...
+    return sympy.ask(
+        sympy.Q.positive(expr), _assume_all(sympy.Q.positive, _free_images_and_symbols(expr))
+    )
+
+
+def _free_images_and_symbols(expr: sympy.Expr) -> Set[sympy.Expr]:
+    return expr.free_symbols | _free_images(expr)
+
+
+def _free_images(expr: sympy.Expr) -> Set[sympy.Expr]:
```
rikardn, line 603: This function only does `set`. Feels to little to warrant a separate function.

src/pharmpy/statements.py
```diff
@@ -593,6 +590,38 @@ def set_lag_time(self, compartment, lag_time):
         return new_comp
 
 
+def _is_positive(expr: sympy.Expr) -> bool:
+    return sympy.ask(
+        sympy.Q.positive(expr), _assume_all(sympy.Q.positive, _free_images_and_symbols(expr))
+    )
+
+
+def _free_images_and_symbols(expr: sympy.Expr) -> Set[sympy.Expr]:
```
rikardn, line 599: Not entierly clear what `images` is referring to.

pharmpy-dev-123: Not sure it is the best name. It is supposed to mean `image` as in https://en.wikipedia.org/wiki/Image_(mathematics).

src/pharmpy/statements.py
pharmpy-dev-123: Yep. We will need a global view of the model if we want to be able to do anything better. However, this change still has value as it fixes an existing issue with the current more naive implementation. Discussing how to improve this even further should not be part of this PR.

src/pharmpy/statements.py
rikardn: What sympy terms `Function`?

src/pharmpy/statements.py
rikardn: What I don't like is to have this many small private function in the top level of statements. It is a bit noisy. Perhaps move them into the function that uses them or merge them into one single function.

src/pharmpy/statements.py
rikardn: Agree

src/pharmpy/statements.py
pharmpy-dev-123: It's nice to have access to both an iterator and a set implementation. It also allows to use Python `yield` syntax without any explicit temporary collection object construction. `_free_images_iter` and `_free_images` end up being simpler to understand without any comment.

src/pharmpy/statements.py
pharmpy-dev-123: It rewrites the right-hand side of an ODE as a weighted sum of `A(i)`s without simplifying the coefficients too much. Blindly simplifying rewrites MM elimination rates which we do not want to happen. I was unable to exploit `UnevaluatedExpr` to achieve the same result.

src/pharmpy/statements.py
pharmpy-dev-123: I think `Function` would be `f`. Here an `image` is `f(x)`.

src/pharmpy/statements.py
pharmpy-dev-123: Yes! It targets `A_i(t)` "images".

rikardn: Ok

src/pharmpy/statements.py
rikardn: Ok, is this form needed in general or only when reading NONMEM models? Does it have to be done when reading the model or can it be done when doing the transformation from explicit system?

src/pharmpy/statements.py
rikardn: To me it adds unnecessary noise since you only needed one of them now. Especially since this is in one of the core modules. If it is only the conversion function that needs these they could be moved inside it.

src/pharmpy/statements.py
rikardn: Ok, sympy calls this `Application`, but that is even less helpful. If the functions are merged or moved into the caller the names could be kept.

src/pharmpy/statements.py
pharmpy-dev-123: I have needed to use it in both `CompartmentalSystem#to_explicit_system` and `pharmpy.plugins.nonmem.advan.compartmental_model`. It is certainly necessary in the former because right-hand sides are generated automatically. Not using it for the latter would require user-provided and pharmpy-generated model sources to follow a more specific format.

rikardn (approved):
Thanks! I'll merge this now.

PR #1107, opened 2022-08-31
Use 'cwres_models' instead of 'models' in resmod results (fixes #911)

### LINKED ISSUE #911: resmod results.csv
Currently, it is difficult to interpret the results.csv as it starts with "models". Would it make sense to follow the same sequence as in the results.csv of the other tools, i.e. summary_tool > summary_models, etc ?

ZheHuang96:
Hi Rikard,

This PR is to distinguish between 'models' and 'summary_models' in resmod.

Best regards,
Zhe Huang


rikardn (approved):
Great!

PR #1112, opened 2022-09-01
Rename 'resmod' to 'ruvsearch' (fixes #912)

### LINKED ISSUE #912: resmod name
Maybe you already through about this, but for structural model development we have "Modelsearch", for IIV "IIVsearch", but for residual error model we have "resmod". This is actually the name of a previously available tool in PsN, which Pharmpy is running now iteratively but it is not exactly the same. Why not calling the current Pharmpy tool "RUVsearch" for the sake of consistency?

ZheHuang96:
Hi Rikard,
There are mainly two points changed

1. 'resmod' has been renamed to 'ruvsearch' 
2. 'ResmodResults' has been renamed to 'RUVSearchResults', this change also impacts psn_resmod because two resmod share the same result class.

Best regards,
Zhe


rikardn:
Some conflicts here now. Also how did you solve the issue with PsN also using the resmod results?

ZheHuang96:
Sorry, I am not sure I understand the issue... What I did is rename this class name  https://github.com/pharmpy/pharmpy/blob/df78da286309864dba46f3e6ac2a7ee4a9727ed1/src/pharmpy/tools/resmod/results.py#L8 to` RUVSearchResults`  then check which file is related to `psn_resmod_results()` , change it  and do the test. Because I am not sure how to generate the .json file like: tests/testdata/nonmem/qa/resmod_results.json

ZheHuang96:
I will check the conflicts, I think they may be related to my last PR.

PR #1116, opened 2022-09-02
Rewrite of rank_models

stellabelin:
Refactoring of `rank_models`, only rank minimized models (option to allow rounding errors), and new attribute `significant_digits` on `ModelfitResults`-class to be able to exclude models where the number of significant digits is unreportable.


stellabelin:
Commit-history messy due to mistake during rebases, this PR should be squashed when merging. Integration tests pass on GHA.

PR #1131, opened 2022-09-06
Skip unsuccessful proxy models in ruvsearch (fixes #1097)

### LINKED ISSUE #1097: Fallback in resmod if iteration model fails
Currently, it can throw the following error: 

```
dofv = base_ofv - model.modelfit_results.ofv
AttributeError: 'NoneType' object has no attribute 'ofv'
```

ZheHuang96:
Hi Rikard,

This PR is to skip the proxy models which didn't have ofvs and only do the comparison between successful proxy models and the base model for each iteration. It would be great if we can have a test for this situation but I feel like it is really hard to have a failed proxy model as an example.

Best regards,
Zhe

### DIFF
--- a/src/pharmpy/tools/ruvsearch/results.py
+++ b/src/pharmpy/tools/ruvsearch/results.py
@@ -1,3 +1,4 @@
+import warnings
 from pathlib import Path
 
 from pharmpy.deps import pandas as pd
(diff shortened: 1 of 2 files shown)


src/pharmpy/tools/ruvsearch/results.py
```diff
@@ -48,29 +49,41 @@ def calculate_results(models):
...
-                        model.modelfit_results.parameter_estimates["sigma_prop"], 6
-                    ),
-                }
-            a = name.split('_')
-            name = '_'.join(a[0:-1])
-
-            model_name.append(name)
-            model_dofv.append(dofv)
-            model_params.append(param)
+            if model.modelfit_results.ofv:
```
rikardn, line 52: Perhaps 

`if model.modelfit_results is not None and model.modelfit_results.ofv is not None:`

rikardn (changes requested):

rikardn:
:+1:

PR #1142, opened 2022-09-07
[dev] Lint lazy dependency imports to prevent import time regressions

pharmpy-dev-123:


rikardn (approved):
Thanks for this! Really useful to not fall into the trap

PR #1145, opened 2022-09-09
Do not print significant digits

ZheHuang96:
Hi Rikard,

@pharmpy-dev-123  and I found the significant digits would be printed after running the tool, e.g. iivsearch, so we thought it should be removed

Best regards,
Zhe


rikardn:
Sure!

Issue #818, opened 2022-05-16
Do not parse coi file when covariance step fail


ZheHuang96:
The issue:
```
distributed.worker - WARNING - Compute Failed
Function:  task
args:      (<Pharmpy model object best_resmod_2>)
kwargs:    {}
Exception: "RuntimeWarning('invalid value encountered in sqrt')"

distributed.worker - WARNING - Compute Failed
Function:  start
args:      (<Pharmpy model object pheno_real>, 4, 0.5, ['time_varying'])
kwargs:    {}
Exception: "RuntimeWarning('invalid value encountered in sqrt')"


tests/integration/test_resmod.py:34 (test_resmod_real)
/home/zhehuang/pharmpy/tests/integration/test_resmod.py:39: in test_resmod_real
    res = run_tool('resmod', model, groups=4, p_value=0.5, skip=[])
/home/zhehuang/pharmpy/src/pharmpy/modeling/run.py:166: in run_tool
    res = execute_workflow(wf, dispatcher=dispatcher, database=database)
/home/zhehuang/pharmpy/src/pharmpy/workflows/execute.py:57: in execute_workflow
    res = dispatcher.run(workflow)
/home/zhehuang/pharmpy/src/pharmpy/workflows/dispatchers/local_dask.py:37: in run
    res = client.get(workflow, 'results')
/home/zhehuang/.local/lib/python3.8/site-packages/distributed/client.py:2722: in get
    results = self.gather(packed, asynchronous=asynchronous, direct=direct)
/home/zhehuang/.local/lib/python3.8/site-packages/distributed/client.py:1977: in gather
    return self.sync(
/home/zhehuang/.local/lib/python3.8/site-packages/distributed/client.py:865: in sync
    return sync(
/home/zhehuang/.local/lib/python3.8/site-packages/distributed/utils.py:327: in sync
    raise exc.with_traceback(tb)
/home/zhehuang/.local/lib/python3.8/site-packages/distributed/utils.py:310: in f
    result[0] = yield future
/home/zhehuang/.local/lib/python3.8/site-packages/tornado/gen.py:762: in run
    value = future.result()
/home/zhehuang/.local/lib/python3.8/site-packages/distributed/client.py:1842: in _gather
    raise exception.with_traceback(traceback)
/home/zhehuang/pharmpy/src/pharmpy/tools/resmod/tool.py:109: in start
    next_res = call_workflow(wf, f'results{current_iteration}')
[...]

PR #1164, opened 2022-09-15
Workaround on pharmpy issue #818

ZheHuang96:
Hi Rikard, 

Here is a temporary workaround for issue #818, covariance matrix, information matrix and correlation matrix will be set to none when the covariance step fails. I have tried some ways I can think to skip parsing files when the covariance step failed but they couldn't work. For me, it is really hard to make them work based on our current class structure, but when we combine the Results class and ChainResults class together, I think we can re-design attribute `_covariate_status` and do the things we really want.

Best regards,
Zhe


rikardn:
Agree. We can merge this so that it doesn't hold up the refactoring.

rikardn (approved):
LGTM! I thing this is ready to merge.

ZheHuang96:
Great, I will solve the conflicts and then merge the PR.

Issue #1054, opened 2022-08-17
Clean up naming of candidate models


stellabelin:
- [x] Save input model in tool database
- [x] #1092
- [x] Save best model name (instead of object) and tool database in results object
- [x] `retrieve_models` should take a results object
- [x] Add function `retrieve_final_model` 
- [x] Use `retrieve_final_model` instead of best model object in AMD-tool
- [x] Remove description between tools in AMD-tool
- [x] Change names/description if needed for subtools

stellabelin:
Relevant issues for first two tasks: #783, #784, and #785

stellabelin:
Input model should be looked over, perhaps stored in tool database rather than as a model database.

stellabelin:
Finished in PR #1242

PR #1167, opened 2022-09-16
Better handling of input and final model in tools

stellabelin:
This includes storing the input model and retrieving final model from tool database instead of having the best model in the results object. Relevant issue: #1054, results from integration tests here (fails on Windows due to same reason as main): https://github.com/pharmpy/pharmpy/actions/runs/3066005603/jobs/4950702949


src/pharmpy/model/results.py
```diff
@@ -134,7 +134,9 @@ def default(self, obj):
             if str(obj.columns.dtype) == 'int64':
                 # Workaround for https://github.com/pandas-dev/pandas/issues/46392
                 obj.columns = obj.columns.map(str)
-            d = json.loads(obj.to_json(orient='table'))
+            # Set double precision to 15 to remove some round-trip errors, however 17 should be set when its possible
+            # See: https://github.com/pandas-dev/pandas/issues/38437
+            d = json.loads(obj.to_json(orient='table', double_precision=15))
```
pharmpy-dev-123, line 157: @rikardn ?

src/pharmpy/results.py
```diff
@@ -5,8 +5,9 @@
 from collections.abc import MutableSequence
 from pathlib import Path
 
-from pharmpy.deps import altair as alt
-from pharmpy.deps import pandas as pd
+import altair as alt
+import pandas as pd
```
pharmpy-dev-123, line 9: Is there a need to change these imports?

tests/integration/conftest.py
```diff
@@ -21,3 +21,16 @@ def start_model(testdata):
         )
         fit(model_start)
     return model_start
+
+
+@pytest.fixture()
```
pharmpy-dev-123, line 26: Maybe add an explicit scope here: https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes

pharmpy-dev-123 (commented):
Some comments. Another idea to retrieve the best model would be to embed sufficient information in the results object so that the best model can be computed from scratch without a name pointer. It would also facilitate revisiting results with different selection criteria, or resuming searching by exploring different branches of the search space.

src/pharmpy/results.py
stellabelin: No, sorry, I think these were accidentally added when I merged the new imports, will change!

tests/integration/conftest.py
stellabelin: Will do!

stellabelin:
What kind of information do you mean? I guess we do have much of the information needed to find the final model (like the summaries with OFV etc.), but some processing would be needed, or the candidate models would need to be read in and ranked again?

pharmpy-dev-123:
> What kind of information do you mean? I guess we do have much of the information needed to find the final model (like the summaries with OFV etc.), but some processing would be needed, or the candidate models would need to be read in and ranked again?

If the results object also embedded the metadata of the search (search space, p-values, etc.) then we could have generic `search_tree`, `search_path`, and `best_model` functions. `best_model` would find the best model given the `Results` object only. `search_path` would give the sequence of models that lead to the best model. `search_tree` would give the entire information as a tree.

Of course, this does not need to be made part of this PR.

src/pharmpy/results.py
pharmpy-dev-123: OK. I see it is resolved now. That was supposed to be caught by `tox -e py38-lint` but the regexes were off. I fixed it now so this does not happen again in the future.

stellabelin:
Okay! Maybe it can be added as an issue?

pharmpy-dev-123:
> Okay! Maybe it can be added as an issue?

Yes. Added #1171.

pharmpy-dev-123 (commented):
The only problem I see in this PR is the precision hack (15 digits vs 17 digits).

stellabelin:
Is it okay for merging? The doc-tests are failing for the same reason as main (which I saw have been fixed).

pharmpy-dev-123:
> Is it okay for merging? The doc-tests are failing for the same reason as main (which I saw have been fixed).

The `lint` step is also failing. Can you rebase on `main` or merge `main` in first and rerun the checks?

rikardn (approved):
I think this can be merged. We'll revisit these things given more experience from the Model class and ModelfitResults rework.

PR #1181, opened 2022-09-21
Make get_rv_parameters more general

stellabelin:


src/pharmpy/modeling/expressions.py
```diff
@@ -928,25 +949,36 @@ def get_pk_parameters(model: Model, kind: str = 'all') -> List[str]:
...
-    free_symbols = set(_pk_free_symbols(cs, kind))
 
-    assignments = list(_remove_synthetic_assignments(classified_assignments))
+def _get_natural_assignments(before_odes):
+    # Return assignments where assignments that are constants (e.g. X=1),
+    # single length expressions (e.g. S1=V), and divisions between parameters
+    # (e.g. K=CL/V) have been filtered out
+    classified_assignments = list(_classify_assignments(list(_assignments(before_odes))))
+    natural_assignments = list(_remove_synthetic_assignments(classified_assignments))
+    return natural_assignments
```
pharmpy-dev-123, line 968: You can also use this better abstraction to refactor `get_individual_parameters`.

src/pharmpy/modeling/expressions.py
```diff
@@ -928,25 +949,36 @@ def get_pk_parameters(model: Model, kind: str = 'all') -> List[str]:
...
+    # Return compartmental system where rates that are synthetic assignments
+    # have been substituted with their full definition (e.g K -> CL/V
+    cs = sset.ode_system.to_compartmental_system()
+
+    assignments = list(_assignments(sset.before_odes))
+    for assignment in reversed(assignments):
+        if assignment not in natural_assignments:
+            # NOTE Substitution must be made in this order
+            cs = cs.subs({assignment.symbol: assignment.expression})
+    return cs
```
pharmpy-dev-123, line 981: It is unfortunate that this now does containment tests on a `list` which is unnecessarily slow.

src/pharmpy/modeling/expressions.py
```diff
@@ -864,22 +864,43 @@ def has_random_effect(model: Model, parameter: str, level: str = 'all') -> bool:
...
 
-    if len(s.expression.free_symbols) > 1:
-        return s.symbol.name
-    else:
-        s = _find_assignment(sset, s.symbol)
-        assert s is not None
-        return s.symbol.name
+    free_symbols = model.statements.free_symbols
+    dependency_graph = _dependency_graph(natural_assignments)
+    filtered = list(_filter_symbols(dependency_graph, free_symbols, {sympy.Symbol(rv)}))
```
pharmpy-dev-123, line 902: Use `sorted` here to get predictable output.

pharmpy-dev-123 (changes requested):
Some things to change or to be discussed.

src/pharmpy/modeling/expressions.py
stellabelin: That's true. Previously I had them in the same function, but felt it was a little messy to return the assignments and the subbed compartmental system from the same function (e.g. the subbed system is not needed for `get_rv_parameters`), and nice to have them separated (maybe clearer what they do when they are in two functions). The list of natural assignments should be relatively short since it's already filtered from the full assignments, not sure how slow this would be.

pharmpy-dev-123: You can always create a temporary `set` object if `Assignment` has a `__hash__` method.

src/pharmpy/modeling/expressions.py
pharmpy-dev-123: You can bypass the `list` construction: `sorted(_filter_symbols(...))`.

src/pharmpy/modeling/expressions.py
```diff
@@ -1080,6 +1112,7 @@ def _classify_assignments(assignments: Sequence[Assignment]):
 
     dependencies = _dependency_graph(assignments)
 
+    # Filter out all symbols that have dependencies (e.g. remove constants X=1)
```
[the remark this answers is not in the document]
stellabelin: Yes, I agree, will change!

rikardn (approved):
This looks good given the earlier comments.

src/pharmpy/modeling/expressions.py
```diff
@@ -857,29 +854,52 @@ def has_random_effect(model: Model, parameter: str, level: str = 'all') -> bool:
...
+        Pharmpy model to retrieve parameters from
+    rv : str
+        Name of random variable to retrieve
 
-    if s is None:
-        raise ValueError(f'Could not find an assignment to {rv}')
+    Return
+    ------
+    Parameters
+        The parameters of the random variable for the given model
```
pharmpy-dev-123, line 878: Return type is `list[str]`. This is also incorrect in the doc string of already present function. I think the idea is that since these are user-facing functions in R they should return reticulatable types.

src/pharmpy/tools/iivsearch/algorithms.py
```diff
@@ -105,7 +105,7 @@ def _is_current_block_structure(etas, combos):
 def _iiv_param_dict(model: Model) -> Dict[str, str]:
     iiv = model.random_variables.iiv
     return {
-        eta: get_rv_parameter(model, eta)
+        eta: get_rv_parameters(model, eta)[0]
```
pharmpy-dev-123, line 108: Maybe it would make sense to have a convenience `_get_rv_parameter` that raises an intelligible error if the number of located parameters is not 1. That could even be a user-facing function. What do you think @rikardn?

stellabelin: Yeah, there should maybe be some check here, or even earlier in input change. However, not sure it need to raise, it is only for description so if a user has the same eta on multiple parameters, the actual tool will still work as expected, it's just the descriptions that will not match with what's happening.

stellabelin: Also I will be rewriting this dict in the PR where I update the description of iovsearch, so maybe this is okay as is for this PR?

pharmpy-dev-123: Yep.

src/pharmpy/modeling/expressions.py
pharmpy-dev-123: @stellabelin ?

stellabelin: Sorry, missed the last comment, updating

src/pharmpy/modeling/expressions.py
stellabelin: Agreed, have updated

src/pharmpy/modeling/expressions.py
stellabelin: Updated

pharmpy-dev-123 (approved):

PR #1182, opened 2022-09-21
Fixes 1101

ZheHuang96:
Hi Rikard,

This PR is to fix issue#1101.

Best regards,
Zhe Huang


CHANGELOG.rst
```diff
@@ -1,3 +1,11 @@
+Next release
+-------------------
+
+Bugfixes
+========
+
+* bd3d868f Set dofv to nan when candidate_model is None (#1101)
```
pharmpy-dev-123, line 7: :warning: Given this is a PR this commit handle might not be the final one.

src/pharmpy/modeling/summarize_individuals.py
```diff
@@ -132,8 +132,14 @@ def ofv(model: Model) -> Union[Series, float]:
...
-def dofv(parent_model: Union[Model, None], candidate_model: Model) -> Union[Series, float]:
-    return np.nan if parent_model is None else ofv(parent_model) - ofv(candidate_model)
+def dofv(
+    parent_model: Union[Model, None], candidate_model: Union[Model, None]
+) -> Union[Series, float]:
+    return (
+        np.nan
+        if parent_model is None or candidate_model is None
+        else ofv(parent_model) - ofv(candidate_model)
+    )
```
pharmpy-dev-123, line 142: LGTM!

CHANGELOG.rst
ZheHuang96: Yes, I see.  I am thinking is there a better way to handle it so that we will not forget to add each fix in the CHAGNELOG? 

> ⚠️ Given this is a PR this commit handle might not be the final one.

pharmpy-dev-123: Put ??? instead?

src/pharmpy/modeling/summarize_individuals.py
pharmpy-dev-123: Actually, if this is meant to fix #1101 I think the problem comes from `model.modelfit_results.individual_ofv` being `None` instead of `np.nan`.

pharmpy-dev-123 (changes requested):

ZheHuang96:
Hi @pharmpy-dev-123 ,

According to the discussion, I set up a test to reproduce the error message. But it is not a 'real' test of the functions `dofv()` and `ofv()` , so I am not sure this test should be kept or removed.

pharmpy-dev-123:
> Hi @pharmpy-dev-123 ,
> 
> According to the discussion, I set up a test to reproduce the error message. But it is not a 'real' test of the functions `dofv()` and `ofv()` , so I am not sure this test should be kept or removed.

I think the test should be changed to test the `ofv` function rather than the `dofv` function.

CHANGELOG.rst
rikardn: It is enough to have the issue number. If the issue #1101 is mentioned here it will have a connection.

ZheHuang96:
> > Hi @pharmpy-dev-123 ,
> > According to the discussion, I set up a test to reproduce the error message. But it is not a 'real' test of the functions `dofv()` and `ofv()` , so I am not sure this test should be kept or removed.
> 
> I think the test should be changed to test the `ofv` function rather than the `dofv` function.

So you want the test for this line:  `np.nan if res is None or res.individual_ofv is None else res.individual_ofv` ? I am not sure whether I can create a dummy model or not, because I think if we want to test the whole functionality, we should have a  model which individual ofv is None.

tests/modeling/test_summarize_individuals.py
```diff
@@ -42,3 +44,10 @@ def test_tflite_not_installed(pheno_path, monkeypatch):
         monkeypatch.setitem(sys.modules, 'tflite_runtime', None)
         df = summarize_individuals([model])
         assert df['predicted_dofv'].isnull().all().all()
+
+
+def test_ofv(pheno_path):
+    parent_model = read_model(pheno_path)
+    candidate_model = Model()
```
pharmpy-dev-123, line 51: Can you also test with a model that has an empty `modelfit_results` object instead of `None`?

pharmpy-dev-123 (changes requested):

tests/modeling/test_summarize_individuals.py
ZheHuang96: So you want a model which is never fitted?

ZheHuang96: Now we have a none modelfit_results:)

pharmpy-dev-123 (approved):
Once unit tests are fixed by @rikardn I think these new tests will pass and it is good to merge.

ZheHuang96:
I will rebase it now! Many thanks!

PR #1211, opened 2022-10-01
[fix] Correct naming in NONMEM plugin

pharmpy-dev-123:
This attempts to get a more deterministic naming for THETAs, ETAs, OMEGAs, EPSs, and SIGMAs.


rikardn:
We will salvage the ideas from this PR, but most likely not the code so closing

PR #1226, opened 2022-10-06
Remove individuals without observations directly by reading nonmem dataset

ZheHuang96:


src/pharmpy/plugins/nonmem/model.py
```diff
@@ -343,19 +343,20 @@ def _read_dataset(self, raw=False, parse_columns=tuple()):
...
-            pass
+        col_names = list(df.columns)
+        have_pk = self.internals.control_stream.get_pk_record()
+        if 'EVID' in col_names:
+            df_obs = df.astype({'EVID': 'float'}).query('EVID == 0')
+        elif 'MDV' in col_names:
+            df_obs = df.astype({'MDV': 'float'}).query('MDV == 0')
+        elif 'AMT' in col_names and have_pk:
+            df_obs = df.astype({'AMT': 'float'}).query('AMT == 0')
         else:
```
rikardn, line 355: Only raise here if have_pk. If not all rows are observations

rikardn (changes requested):

src/pharmpy/plugins/nonmem/model.py
ZheHuang96: Thanks, I will change it.

ZheHuang96: > From the NONMEM documentation it seems as if EVID is also only interpreted if $PK. MDV is unclear from the documentation, but seems to be general. Could be tested in a resmod model.

Yes, this is also my question, because I check the `parse_datainfo()` function https://github.com/pharmpy/pharmpy/blob/37144d60b0d988822c37b7f1462d6a071eb71200/src/pharmpy/plugins/nonmem/parsing.py#L610 I think our current way is checking the PK record for EVID, MDV and AMT. I think maybe we should apply same strategy here. I can have a test on the resmod model to see what happens.

rikardn: Great. If the MDV is not used in a $PRED model then the default should be to have all records as observations and only consider EVID, MDV and AMT if model has $PK

ZheHuang96: Here is the result: ` (WARNING  13) WITH USER-WRITTEN PRED OR $PRED, NM-TRAN CANNOT APPEND THE
 MDV DATA ITEM.`
I think this warning indicates the MDV is not used in the $PRED model?

rikardn:
Thanks for the update. I think this is ready to merge

rikardn (approved):

PR #1231, opened 2022-10-07
[deps] Exclude pywin32 versions 301 and 304 because of incompatibilities with conda

pharmpy-dev-123:
Q: How do we ensure this is not reverted by dependabot?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -78,6 +78,7 @@ pyreadr==0.4.7
 pyrsistent==0.18.1
 python-dateutil==2.8.2
 pytz==2022.4
+pywin32!=301,!=304; sys_platform == 'win32' and platform_python_implementation != 'PyPy'
 PyYAML==6.0
 pyzmq==24.0.1
 requests==2.28.1


rikardn:
Why wasn't pywin in requirements.txt before? Is it a new dependency?

rikardn:
I am asking because we also have a script to update requirements.txt so that it contains all dependencies of what is required from setup.py

pharmpy-dev-123:
> Why wasn't pywin in requirements.txt before? Is it a new dependency?

I think it is because it is a Windows-only dependency. Currently it is explicitly required by one of our explicit requirements, `jupyter-sphinx`, via `jupyter-core`. It is used by `appdirs` 1.4.4, and that's were the error is raised from, but `appdirs` does not have an explicit list of requirements, it just assumes `pywin32` is installed.

pharmpy-dev-123:
See https://github.com/jupyter/jupyter_core/blob/5f9e8ebff5857ec41c7d4ed9727051bb00520ac8/pyproject.toml#L22. Maybe we could add the `!= "PyPy` condition also.

rikardn:
Aha. Sneaky dependency. Then we'll tackle issues with dependabot and the update script if they arise later.

pharmpy-dev-123:
Just for the record: This is one of the many traces @ZheHuang96 and I got trying to run Pharmpy's tests through conda

```
ImportError while loading conftest 'tests\integration\conftest.py'.
tests\integration\conftest.py:7: in <module>
.tox\integration\lib\site-packages\pharmpy\model\__init__.py:1: in <module>
    from .data import DatasetError, DatasetWarning
.tox\integration\lib\site-packages\pharmpy\model\data.py:24: in <module>
    import pharmpy.config as config
.tox\integration\lib\site-packages\pharmpy\config.py:50: in <module>
    config_file = read_configuration()
.tox\integration\lib\site-packages\pharmpy\config.py:40: in read_configuration
    user_path = user_config_dir()
.tox\integration\lib\site-packages\pharmpy\config.py:19: in user_config_dir
    user_path = Path(appdirs.user_config_dir(appname)) / configuration_filename
.tox\integration\lib\site-packages\appdirs.py:196: in user_config_dir
    path = user_data_dir(appname, appauthor, None, roaming)
.tox\integration\lib\site-packages\appdirs.py:81: in user_data_dir
    path = os.path.normpath(_get_win_folder(const))
.tox\integration\lib\site-packages\appdirs.py:480: in _get_win_folder_with_pywin32
    from win32com.shell import shellcon, shell
E   ImportError: DLL load failed while importing shell: The specified procedure could not be found.
ERROR: InvocationError for command '.tox\integration\Scripts\pytest.EXE' -vv 'tests\integration\test_allometry.py' (exited with code 4)
```

pharmpy-dev-123:
See for instance https://github.com/conda/conda/issues/11503 for the same problem happening in the wild.

pharmpy-dev-123:
@rikardn This is needed. Was difficult to run `git bisect` with Windows/miniconda on a branch that did not have this commit. It does nothing on non-Windows platforms, and it can be removed if we drop the indirect dependency on `jupyter-core`.

PR #1232, opened 2022-10-07
Copy all files from tmp directory when nonmem run function crash

### LINKED ISSUE #1082: Copy tmp run directory if tool crashes
E.g. if a tool fails due to timeout or a raised error, Pharmpy should copy the NONMEM directory from tmp so that it doesn't get deleted. Check if Dask already has system for this

ZheHuang96:
Working on finding a failed run example to test. Fixes #1082.


rikardn:
Closing in favour of #1352

PR #1234, opened 2022-10-07
Set zero_protection=True as default in set_proportional_error_model (fixes #969)

### LINKED ISSUE #969: Apply zero-protect by default
Hi, 
The initial_model (created through combination of assemblerr model and pharmr function) did not start for the first dataset tested (IV route).  We experienced the same problem with a second dataset for IV route again
When applying the zero_protect function, the initial_model run well.  We still need to test the zero_protect function with the second dataset and see if it solves the issue as for the first dataset.

Our suggestion would be to applying by defaut the zero_protect function for PROP RUV whatever the route and not only for oral route as it currently done.

ZheHuang96:
Hi Rikard, 

This PR is to fix #969 by setting zero_protection = True as default.

Best regards,
Zhe

### DIFF
--- a/src/pharmpy/modeling/error.py
+++ b/src/pharmpy/modeling/error.py
@@ -162,7 +162,7 @@ def _get_prop_init(model):
         return (dv_min / 2) ** 2
 
 
-def set_proportional_error_model(model, data_trans=None, zero_protection=False):
+def set_proportional_error_model(model, data_trans=None, zero_protection=True):
     r"""Set a proportional error model. Initial estimate for new sigma is 0.09.
 
     The error function being applied depends on the data transformation.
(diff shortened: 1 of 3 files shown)


pharmpy-dev-123:
Changes look good to me, but does this address all the questions raised in the issue? Specifically, the first question was about RUVsearch. I do not know enough about what tools use what modeling functions to understand whether this change is sufficient.

ZheHuang96:
Morning@pharmpy-dev-123, I think the issue is related to generating the start model of AMD process. I think previously when they created the start model, they needed to make it by the combination of assemblerr model and pharmpy function. It seems the suggestion from our side is that only set `zero_protection=True`  for `set_proportional_error_model` when it is an oral dataset. Because modelsearch tool needs it. But during the tests, some IV datasets also needed zero protection for modelsearch tool. Now, if the start model is generated automatically by the AMD function in pharmpy, it will have zero protection whatever the dataset type is. 

This PR just follows their suggestion to set `zero_protection = True` as default. I guess it will be more convenient for them to use `set_proportional_error_model` since the majority of situations of the `zero_protection` is True.

rikardn (approved):
LGTM

PR #1235, opened 2022-10-07
Update tox code in Contribute

ZheHuang96:
Hi,

I accidentally find the tox code in Contribute needs to be updated.

### DIFF
--- a/docs/contribute.rst
+++ b/docs/contribute.rst
@@ -83,7 +83,7 @@ Check code formatting
 
 Pharmpy use the ``black`` code formatter, the ``flake8`` linter and ``isort`` for formatting of imports. Formatting and linting can be run separately with::
 
-    tox -e check
+    tox -e format
 
 Developers can format the code any way they want while writing code and the formatter will automatically reformat.
 
@@ -96,11 +96,11 @@ To run all unit tests for one python version::
 
 To run a particular unit test file::
 
-    tox -e py39 -- pytest test_parameters.py
+    tox -e py39 -- tests/modeling/test_parameters.py
 
 To run a particular test function in a particular test file::
 
-    tox -e py39 -- pytest test_parameters.py::test_myfunc
+    tox -e py39 -- tests/modeling/test_parameters.py::test_get_thetas
 
 Build and test the documentation
 ********************************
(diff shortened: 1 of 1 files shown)


docs/contribute.rst
```diff
@@ -130,7 +130,7 @@ To run all integration tests::
 
 To run a particular test function in a particular integration test file::
 
-    tox -e integration -- pytest tests/integration/test_modelsearch.py -k test_summary_individuals
+    tox -e integration -- tests/integration/test_modelsearch.py -k test_summary_individuals
```
pharmpy-dev-123, line 133: Use `::` instead of `-k` here also: `tox -e integration -- tests/integration/test_modelsearch.py::test_summary_individuals`

pharmpy-dev-123 (changes requested):
Just one small thing to change. Otherwise, LGTM!

rikardn (approved):
Good catches!

docs/contribute.rst
rikardn: Yes this might be better. Does `::` do exact matches instead of pattern match as `-k`?

ZheHuang96: Just have a quick test by running:
 `tox -e integration -- tests/integration/test_modelsearch.py::test_summary_individuals` and 
`tox -e integration -- tests/integration/test_modelsearch.py -k test_summary_individuals`
Here is the diff:
```4,5c4,5
< integration run-test-pre: PYTHONHASHSEED='587217178'
< integration run-test: commands[0] | pytest -vv tests/integration/test_modelsearch.py::test_summary_individuals
---
> integration run-test-pre: PYTHONHASHSEED='1911634784'
> integration run-test: commands[0] | pytest -vv tests/integration/test_modelsearch.py -k test_summary_individuals
12c12
< collected 1 item                                                                                                                                                                                                              
---
> collected 9 items / 8 deselected / 1 selected                                                                                                                                                                                 
```
When `-k`,  tox will collect all items in the file and select the one we want while `::`, tox will directly collect the item we want.

rikardn: Ok, that is another difference. Then `::` is the way to go in our contribution doc.

PR #1247, opened 2022-10-13
Remove some places where database attribute on model is still set

stellabelin:
Add fix-me comments in remaining places

### DIFF
--- a/src/pharmpy/modeling/common.py
+++ b/src/pharmpy/modeling/common.py
@@ -21,7 +21,6 @@ from pharmpy.model import (
     RandomVariables,
 )
 from pharmpy.utils import normalize_user_given_path
-from pharmpy.workflows import default_model_database
 
 
 def read_model(path):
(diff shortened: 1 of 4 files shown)


src/pharmpy/modeling/common.py
```diff
@@ -128,6 +127,7 @@ def read_model_from_string(code, path=None):
         path = normalize_user_given_path(path)
         import pharmpy.workflows
 
+        # FIXME model.database should be removed
         model.database = pharmpy.workflows.default_model_database(path)
```
rikardn, line 131: What is the reason for keeping this in?

rikardn: Could it be removed already now?

stellabelin: Should be okay! But then I guess the `path` argument doesn't make sense atm?

rikardn: I think path can be removed. Since it is not needed for the database any longer it wouldn't be needed to parse a model from a string.

stellabelin: Fixed!

rikardn: Nice!

rikardn (approved):
I think this can be merged

PR #1248, opened 2022-10-13
Bugfix and amd example

ZheHuang96:
Hi,

This PR includes:

- Use the (nonzero-minimum-time)/2 as the initial estimates for MDT and MAT, which we discussed yesterday.
- Roll back to the full NLME model of the previous iteration when the full NLME model fails in the current iteration in ruvsearch. E.g. if `best_ruvsearch_2` fails, `best_ruvsearch_1` will be the output model;, if `best_ruvsearch_1` fails, input model(start model) will be the output model.
- Add simulated dataset and datainfo files for the AMD test and documentation. I didn't change the search space of modelsearch. I have compared the model selections between our current search space(P1) with a larger search space(P1-LAG), there are two benefits when using P1 as the search space: 
  - It could pick up a model from covsearch tool while the P1-LAG model did the covsearch and didn't select any model. Since the test result is actually for AMD documentation, I think maybe it is much better to show for each step the tool can pick up a model. 
  - It has less run time so that the total run time of the integration test won't increase too much on GHA.
- Update amd_results.json for AMD documentation.

Best regards
Zhe


rikardn (approved):
LGTM

ZheHuang96:
Hi! I will merge it. This dataset is only for the AMD documentation and test because it only simulated DV values. I am generating another simulated dataset which includes pseudo covariates.

PR #1261, opened 2022-10-18
[deps] Upgrade pandas to 1.5.0

### LINKED ISSUE #1223: Fix issues with pandas 1.5.0
See  #1177

pharmpy-dev-123:
Needed some changes due to FutureWarnings. Some could be fixed, others had to be filtered out because of upstream usage in altair.

Note that all changes are compatible with the current `setup.py` constraint on `pandas` (`>=1.4`).

Replaces #1177. Fixes #1223.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -60,7 +60,7 @@ notebook==6.5.1
 numexpr==2.8.3
 numpy==1.23.4
 packaging==21.3
-pandas==1.4.4
+pandas==1.5.0
 pandocfilters==1.5.0
 parso==0.8.3
 partd==1.3.0
--- a/src/pharmpy/math.py
+++ b/src/pharmpy/math.py
@@ -56,7 +56,7 @@ def round_and_keep_sum(x, s):
     """
     sorted_fractions = x.apply(lambda x: math.modf(x)[0]).sort_values(ascending=False)
     rounded_sample_sizes = x.apply(lambda x: math.modf(x)[1])
-    for (group_index, _) in sorted_fractions.iteritems():
+    for (group_index, _) in sorted_fractions.items():
         num_samples = rounded_sample_sizes.sum()
         diff = s - num_samples
         if diff == 0:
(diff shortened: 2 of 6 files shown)


tests/test_cli.py
```diff
@@ -99,6 +99,10 @@ def test_add_iov(datadir, options, tmp_path):
         assert re.search(r'ETAI1', mod_cov)
 
 
+@pytest.mark.filterwarnings(
```
rikardn, line 102: Should the filtering be done in the actual code triggering it or would that be sweeping it under the carpet for the user?

tests/tools/test_bootstrap.py
```diff
@@ -1,9 +1,14 @@
 import pandas as pd
+import pytest
 
 from pharmpy.results import ModelfitResults, read_results
 from pharmpy.tools.bootstrap.results import calculate_results
 
 
+@pytest.mark.filterwarnings(
+    # FIXME Remove filter once altair stops relying on deprecated APIs
```
rikardn, line 9: Should we perhaps open an issue on Altair?

tests/test_cli.py
pharmpy-dev-123: I think you are correct that it would be better user-wise to filter it in the actual code.

tests/tools/test_bootstrap.py
pharmpy-dev-123: Yes. I'll double check if it's an `altair` issue or `vega-lite` issue.

pharmpy-dev-123: Definitely `altair`. Will open an issue.

tests/test_cli.py
pharmpy-dev-123: Changed!

rikardn (approved):
Looks good!

tests/tools/test_bootstrap.py
pharmpy-dev-123: See https://github.com/altair-viz/altair/issues/2698.

PR #1275, opened 2022-10-20
Test amd with large search space

ZheHuang96:
Hi Stella,

This PR is based on our discussion of reasonable search space. So here is the new change and except RUVSearch, other tools can select a candidate model. Maybe we should try it on CLI?

Also, do I need to update the code in amd documentation?

Best regards,
Zhe


rikardn:
How much will this increase the test time?

stellabelin:
Hi! I don't think we need the bigger search space when testing, to me it seems like enough to have it when we generate the AMD-results file for documentation. The code in the documentation reads this in already so no changes needed there, but perhaps it would be good in the documentation to mention the input arguments used?

ZheHuang96:
Ok, great! Then let us keep the original example. and yes, I think we need to specify the input options in the results part of amd documentation so that it can be distinguished from the example code.
By the way, just want to double-check, is the original example fine for CLI?

PR #1283, opened 2022-10-24
[fix] Make dataset deduplication independent of Python's hash seed

pharmpy-dev-123:
This implementation takes data, dtypes, columns, and index into account for hash computation.

### DIFF
--- a/src/pharmpy/internals/df.py
+++ b/src/pharmpy/internals/df.py
@@ -1,6 +1,31 @@
+from __future__ import annotations
+
+from hashlib import sha256
+from typing import Iterator, Union
+
 from pharmpy.deps import pandas as pd
 
 
-def hash_df(df) -> int:
-    values = pd.util.hash_pandas_object(df, index=True).values
-    return hash(tuple(values))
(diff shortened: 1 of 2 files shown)


src/pharmpy/internals/df.py
```diff
@@ -1,6 +1,31 @@
...
+
+
+def _df_hash_values(df: pd.DataFrame) -> Iterator[pd.Series]:
+    yield _pd_hash_values(df.columns)
+    yield _pd_hash_values(df.index)
+    yield _pd_hash_values(df.dtypes)
+    yield _pd_hash_values(df)
+
+
+def hash_df_runtime(df: pd.DataFrame) -> int:
```
stellabelin, line 23: Is this used anywhere?

stellabelin (commented):
LGTM!

src/pharmpy/internals/df.py
pharmpy-dev-123: No, but I think it could be used if we ever need to put dataframes in sets or dicts.

rikardn (approved):
LGTM

PR #1304, opened 2022-10-31
[perf] Reduce parsing complexity of NONMEM grammars 

pharmpy-dev-123:
All grammars are now LALR(1). Needed to exploit '%ignore' for that to work. Lark usage is down to <2% (was 58%).


stellabelin (commented):
How come there are extra `\n`-characters in the tests? Does the code itself look the same as previously or are there extra newlines? Otherwise LGTM!

pharmpy-dev-123:
> How come there are extra `\n`-characters in the tests? Does the code itself look the same as previously or are there extra newlines? Otherwise LGTM!

I should refactor and add those conditionally in the test to avoid changing so many lines. Ideally they would not be necessary, but it's much easier for the parser to know that each line must end with a newline. I still need to discuss about this with @rikardn.

PR #1317, opened 2022-11-02
[feature] Move AMD's covariate search space definition into search_space parameter

pharmpy-dev-123:
> @rikardn Ready to MERGE

  - [x] Filter AMD MFL string before passing down to subtools
  - [x] Pass search spaces as MFL string down to subtools
  - [x] Reject unknown MFL features in subtools (in `validate_input`)
  - [x] Update docs
  - [x] Integration tests pass


src/pharmpy/tools/amd/run.py
```diff
@@ -111,14 +106,21 @@ def run_amd(
...
+                'COVARIATE(@IIV, @CATEGORICAL, cat, *)'
             )
         else:
-            search_space = 'ELIMINATION([MM,MIX-FO-MM]);' 'PERIPHERALS([1,2])'
+            search_space = (
+                'ELIMINATION([MM,MIX-FO-MM])\n'
+                'PERIPHERALS([1,2])\n'
+                'COVARIATE(@IIV, @CONTINUOUS, exp, *)\n'
+                'COVARIATE(@IIV, @CATEGORICAL, cat, *)'
+            )
```
stellabelin, line 124: Should there be defaults if the user only writes a search space for the structural model or vice versa? Previously, we set defaults for both and you could omit one of them to get a sensible default.

stellabelin (changes requested):
Looks good but see the comment about search space defaults. Also, the documentation for AMD needs to be updated!

src/pharmpy/tools/amd/run.py
pharmpy-dev-123: That's a good idea. First solution I can think of is to have a way to decide whether a given MFL string contains any particular directive. For instance if there are no `COVARIATE` directives, then we should add the default ones.

stellabelin: Yes, that sounds good! And if you only have `COVARIATE` directives we assume that no search space have been provided? Or maybe specify in respective module which features are included (now we could maybe use covariate vs non-covariate but I was thinking if this is expanded)?

pharmpy-dev-123: Each MFL statement has a type, so we can just have a list of types and if no statement has one of these types we add some default ones.

pharmpy-dev-123: Done!

pharmpy-dev-123:
> Looks good but see the comment about search space defaults. Also, the documentation for AMD needs to be updated!

Documentation updated.

rikardn (approved):
This is looking good. I think we can merge

pharmpy-dev-123:
Just rebased. There were conflicts.

PR #1352, opened 2022-11-16
Handle missing results nonmem

stellabelin:


rikardn (approved):
LGTM

src/pharmpy/plugins/nonmem/run.py
```diff
@@ -72,10 +73,16 @@ def execute_model(model, db):
...
-    except FileNotFoundError:
-        pass  # NOTE warns later when looking for .lst and .ext file
+    results_path = model_path / 'results.lst'
+    start = time.time()
+    timeout = 5
+
+    while time.time() - start < timeout:
+        if results_path.is_file():
+            results_path.rename((model_path / basename).with_suffix('.lst'))
+            break
```
pharmpy-dev-123, line 83: Better to keep `try`-`except` here.

pharmpy-dev-123: Exactly that error. `rename` will already do the check for you somehow and raise if there is a problem. Anything could happen between the `is_file` and `rename` calls. I agree it's unlikely to cause a problem here, but it's better to adopt that code style for IO. See for instance https://docs.quantifiedcode.com/python-anti-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html.

It is especially important to adopt that kind of code style for IO (file system, network, etc.) because of the potential for race conditions (which can be safely ignored for in-memory stuff as we assume errors will not happen there).

pharmpy-dev-123: See for instance this more interesting FS example: https://stackoverflow.com/questions/21517227/how-to-avoid-a-race-condition-with-makedirs

rikardn: `try`-`except` seems safer and I take it we don't need both since there is a specific `FileNotFound` exception.

stellabelin: Okay, I'll change to `try`-`except`, do we need one for e.g. `PermissionError` as well?

rikardn: So the question is what to do for any other error than `FileNotFound`. Now any such error would `raise` from this function. Do we still want to wait for the timeout, i.e. handle any `IOException` in the same way as `FileNotFound`? That might require changes to later code as well. Perhaps we can assume that the only thing that could happen while renaming a file in /tmp is that it cannot be seen yet and if other problems arise in production we fix it then.

Issue #1139, opened 2022-08-29
Length of values (288) does not match length of index (296)


joaoabr:

rikardn:
To be able to figure this one out we would need the ID, MDV and EVID columns of one of the datasets. Would it be possible to send that to us?

joaoabr:
I have just shared via Box the file modelsearch_candidate2.csv with columns ID, AMT, DV and EVID. AMT and DV have been replaced by dummy values if >0. Let me know if you need more.

ZheHuang96:
This issue is because `len(nobsi.index)` and `len(ids)` are not match, there are 8 individuals has no observations, they only have dose records.

ZheHuang96:
Hi @joaoabr , I think it would be great if you can also give us the dummy dataset 12, you can apply the same strategy as you did for modelsearch_candidate2.csv. We would like to make sure the fix can handle all the cases that you reported. 

Many thanks!

rikardn:
@ZheHuang96 so you found som individuals without observations. How did they slip past the filtering?

ZheHuang96:
Yes, in the dataset of model 5, there are 8 subjects without observations. I was wondering where do we filter the dataset? I remember you showed me last week, but I forgot about it. Is it the  `read_dataset()` in plugins/nonmem/dataset.py? I looked through the code but I couldn't find it.

And the second case that @joaoabr posted showed the opposite situation, I think it could be a precious case for us to understand what happens.

rikardn:
The filtering is in `_read_dataset` in the nonmem `model.py`. Feels like it shouldn't be here.

So since we are doing filtering how can there be a difference between `get_ids` and the observations index?

ZheHuang96:
> The filtering is in `_read_dataset` in the nonmem `model.py`. Feels like it shouldn't be here.
> 
> So since we are doing filtering how can there be a difference between `get_ids` and the observations index?

I can have a quick test.

joaoabr:
@ZheHuang96 I have just shared via Box the file modelsearch_candidate2.csv with columns C, ID, AMT, RATE, TIME, DV and EVID. AMT, RATE and DV have been replaced by dummy values if >0. Let me know if you need more.

ZheHuang96:
I can reproduce the first case: ` ValueError('Length of values (288) does not match length of index (296)')`, there are 8 individuals without any observations, we should filter out these individuals but I think there might be an error in your dataset/model so it didn't behave as we expect.   
[...]

PR #1356, opened 2022-11-17
Add format option to TABLE if ID > 99999

stellabelin:
This solves part of #1139 (long IDs get truncated).

### DIFF
--- a/tests/nonmem/test_nonmem_model.py
+++ b/tests/nonmem/test_nonmem_model.py
@@ -24,6 +24,7 @@ from pharmpy.modeling import (
     add_population_parameter,
     create_joint_distribution,
     remove_iiv,
+    set_estimation_step,
     set_initial_estimates,
     set_zero_order_absorption,
     set_zero_order_elimination,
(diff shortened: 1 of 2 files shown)


tests/nonmem/test_nonmem_model.py
```diff
@@ -1007,6 +1008,27 @@ def test_convert_model(testdata):
...
+    $OMEGA 2 ; OM1
+    $SIGMA 3 ; SI1
+    $ESTIMATION METHOD=1 INTER
+    """
+    model = Model.create_model(StringIO(code))
+    dataset_new = model.dataset.copy()
+    dataset_new['ID'] = dataset_new['ID'] * 10000
+    model.dataset = dataset_new
+    set_estimation_step(model, 'FO', residuals=['CWRES'])
+    assert 'FORMAT=' in model.model_code
```
pharmpy-dev-123, line 1029: Would it make sense to also check that adding `FORMAT` actually solves the issue?

stellabelin: I think the only way to test if it solves the issue is by running this in NONMEM, reading the output table and making sure we have an expected numbers of IDs, I guess we could add an integration test but maybe it's enough to see that Pharmpy adds it when appropriate? Not sure

Issue #816, opened 2022-05-12
Move modeling-functions that run tools to tool-module


stellabelin:
To avoid circular imports. Need to consider consequences in R (if importing both `modeling` and `tools`, how to avoid clashing names etc.) and documentation.

pharmpy-dev-123:
Fixed in #838 (295782f0).

PR #1358, opened 2022-11-17
[DRAFT] Create pharmpy.tools.run_* functions on demand

pharmpy-dev-123:
Related to #816.


pharmpy-dev-123:
Was superseded by better implementation that's already on `main`. Closing.

PR #1375, opened 2022-11-24
[fix] Ignore dask-worker-space cleanup errors on Windows in dispatcher

pharmpy-dev-123:
> @rikardn Ready to MERGE

  - [x] Integration tests: https://github.com/pharmpy-dev-123/pharmpy/actions/workflows/integration.yml?query=branch%3Afix-workflows-local-dask-dask-worker-space-dirlock-1

### DIFF
--- a/src/pharmpy/internals/fs/tmp.py
+++ b/src/pharmpy/internals/fs/tmp.py
@@ -1,83 +1,104 @@
-import os
-import shutil
-import time
-import warnings
-import weakref
+import os as _os
+import shutil as _shutil
+import warnings as _warnings
+import weakref as _weakref
 from tempfile import mkdtemp
 
-if os.name == 'nt':
-    # This is copied and modified from the python 3.9 implementation
-    # The aim is to be able to handle Permission issues in Windows
-    class TemporaryDirectory:
-        """Create and return a temporary directory.  This has the same
-        behavior as mkdtemp but can be used as a context manager.  For
-        example:
 
-            with TemporaryDirectory() as tmpdir:
-                ...
+# NOTE This is copied from the Python 3.10 implementation to expose the
+# ignore_cleanup_errors flag and patch PermissionError handling for Windows
+# lock files. Here is the original source link:
+#
+# https://github.com/python/cpython/blob/b494f5935c92951e75597bfe1c8b1f3112fec270/Lib/tempfile.py#L778-L848
+#
+# We had to remove the definition
+# __class_getitem__ = classmethod(_types.GenericAlias)
+# introduced in Python 3.9 because we do not use it and having a correct
+# implementation for Python 3.8 would be non-trivial. This definition can be
+# reintroduced if needed once we drop support for Python 3.8.
+#
+# The whole thing can be replaced by `from tempfile import TemporaryDirectory`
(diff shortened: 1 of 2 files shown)


src/pharmpy/internals/fs/tmp.py
```diff
@@ -1,83 +1,92 @@
...
+        return self.name
 
-        def cleanup(self):
-            if self._finalizer.detach():
-                self._rmtree(self.name)
+    def __exit__(self, exc, value, tb):
+        self.cleanup()
 
-else:
-    from tempfile import TemporaryDirectory  # noqa
```
pharmpy-dev-123, line 83: Using the same implementation on Windows, Linux, and MacOS makes it simpler to do what we want without having to write too much extra code. Otherwise we would have to have some additional code conditional on `os.name` to figure out if we need to pass the `ignore_cleanup_errors` flag. Here we can just use `TemporaryDirectory` as if it was coming from Python 3.10. There has not been any change to the source since then, so the implementation is exactly the same in Python 3.11. Moreover, I guess that if newer versions introduced important patches we would likely need to add those to our copy anyway. Once support for Python 3.8 and 3.9 is dropped, we can just discard the whole thing.

src/pharmpy/workflows/dispatchers/local_dask.py
```diff
@@ -12,41 +13,63 @@
...
-            dsk = workflow.as_dask_dict()
+    # NOTE We change to a new temporary directory so that all files generated
+    # by the workflow end-up in the same root directory. Each task of a
+    # workflow has the responsibility to avoid collisions on the file system
+    # (using UUIDs as filenames for instance). This also allows tasks to share
+    # files on the local file-system if needed. Since in the future we may want
+    # to run tasks on remote machines it might be a good idea to instead not
+    # assume tasks share the same working directory or even filesystem, and
+    # let them communicate through some other mean and create their own
+    # temporary directory if needed.
```
pharmpy-dev-123, line 24: I guess tasks would not even be able to use `chdir` because they could be run on threads and that would not be nice.

src/pharmpy/internals/fs/tmp.py
```diff
@@ -1,83 +1,104 @@
...
                         # PermissionError is raised on FreeBSD for directories
-                        except (IsADirectoryError, PermissionError):
-                            time.sleep(0.1)
-                            cls._rmtree(path)
-                    except FileNotFoundError:
-                        pass
-                elif issubclass(exc_info[0], FileNotFoundError):
+                        # and by Windows on lock files used by other processes
+                        try:
+                            cls._rmtree(path, ignore_errors=ignore_errors)
```
pharmpy-dev-123, line 71: This somehow assumes a `PermissionError` will not be raised again, which is bold, since it can potentially recurse forever. This is not a problem we have encountered, and it is taken straight from `cpython`'s source so I did not change it.

src/pharmpy/workflows/dispatchers/local_dask.py
pharmpy-dev-123: Updated.

rikardn:
LGTM!

PR #1510, opened 2023-01-26
Verification of conversion from nonmem to nlmixr2

JohanWBorg:
New function in pharmpy.plugin.nlmixr.model called verification(). Takes a nonmem model and a tool database name as input. Convert nonmem to nlmixr and compute predicted values to be compared to that of the nonmem model. If all predictions are within the margin of error (default to 10^-3), return True, otherwise False. If argument return_comp is True, instead return pd.dataframe with comparison of predictions from both models. Also added function modify_dataset() which takes in a model and returns a model with "evid" added to the dataset (mandatory for runnning nlmixr).

Some changes have also been made to convert_model() in order to achieve the wanted predictions and comparison.


rikardn:
Thanks for the PR! This looks good to me.

PR #1662, opened 2023-04-14
Update to nlmixr conversion tool

JohanWBorg:
Update to nlmixr conversion tool for greater error model handling and handling of more limitations

### DIFF
--- a/src/pharmpy/plugins/nlmixr/model_block.py
+++ b/src/pharmpy/plugins/nlmixr/model_block.py
@@ -65,8 +65,8 @@ def add_statements(
                 if s.expression.is_Piecewise:
                     convert_piecewise(s, cg, model)
                 else:
-                    expr, error = find_term(model, s.expression)
-                    add_error_model(cg, expr, error, s.symbol.name)
+                    expr, error = find_term(model, s.expression, cg)
+                    add_error_model(model, cg, expr, error, s.symbol.name)
                     add_error_relation(cg, error, s.symbol)
                     error_model_found = True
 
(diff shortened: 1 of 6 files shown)


src/pharmpy/plugins/nlmixr/sanity_checks.py
```diff
@@ -62,9 +61,24 @@ def check_model(
         print_warning("Omega with value same not supported. Updated as follows.")
         model = change_rvs_same(model, omega=True)
 
+    # Checks regarding esimation method
+    if method := known_estimation_method(model):
```
rikardn, line 65: This is confusing. The function will return the name of an unknown estimation method and `None` if it is known. Perhaps this could be made clearer

src/pharmpy/plugins/nlmixr/model_block.py
```diff
@@ -101,15 +133,98 @@ def add_statements(
...
+
+    # Assume all sigma with a fixed inital value of 1 to be removed
+    # TODO: Remove when using res_error_term instead
+    from pharmpy.modeling import get_sigmas
+
+    sigma_to_remove = set()
+    for sigma in get_sigmas(model):
+        # TODO : also remove alias for sigma
+        if sigma.init == 1 and sigma.fix:
+            cg.remove(f"{sigma.name} <- fixed({sigma.init})")
```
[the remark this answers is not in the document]
JohanWBorg: Yes, this has now been changed. All sigmas that are fixed to one are not added now. In the future, these should be removed if necessary when analyzing the error model.

src/pharmpy/plugins/nlmixr/sanity_checks.py
JohanWBorg: I agree, the function has been changed to be a bit clearer

rikardn (approved):

Issue #1569, opened 2023-03-08
Bug when parsing $DES


JohanWBorg:
### Bug report

using pharmpy version 0.91.0
When reading a model with the structure below, only the first two ODEs in the $DES block are interpreted as differential equations while the remaining two are left out.
```
$MODEL
COMP=FIRST
COMP=SECOND
COMP=THIRDS
COMP=FOURTH
...
$DES
DADT(1) = K1*A(2)*(A(3)+P1) - (K2+KOUT)*A(1) - KSEC*A(4)*A(1)
DADT(2) = KIN + K2*A(1) - K1*A(2)*(A(3)+P1)  - KOUT*A(2)
DADT(3) =-K*A(3)
DADT(4) = SL
```

PR #1682, opened 2023-04-24
Correction of ODE parsing and $DES assignment statements

JohanWBorg:
Should solve issue  #1568 (There seem to have been a duplication with issue #1569)


rikardn:
LGTM

PR #1705, opened 2023-05-02
Bioavailability fixes and updates to conversion tool

JohanWBorg:
Fixed issue with get_bioavailability not working as intended. It was mainly due to:
- When comparing two compartments, bioavailability was not taken into account and hence nodes were not updated as wanted when setting the bioavailability
- When setting dose/bioavailability/alag, the "comp" variable used was not updated and hence did not match the nodes in the nx graph. Added function find_compartment to the CompartmentalSystemBuilder class to avoid building the compartment system each time a compartment is needed. Please let me know if this should be done in some other way

Also included some updates to the conversion tool such as
- Bioavailability conversion
- Further developed error model
- Added support for sigma correlation in RxODE conversion
- Fixed incorrect handling of parameter estimates in nlmixr2 execution
- Removed some unnecessary redundancies

### DIFF
--- a/src/pharmpy/model/statements.py
+++ b/src/pharmpy/model/statements.py
@@ -390,6 +390,11 @@ class CompartmentalSystemBuilder:
         mapping = {compartment: new_comp}
         nx.relabel_nodes(self._g, mapping, copy=False)
         return new_comp
+    
+    def find_compartment(self, name):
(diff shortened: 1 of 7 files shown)


rikardn:
LGTM!

PR #1774, opened 2023-06-12
Update to nlmixr2/Rxode2 conversion tool

JohanWBorg:
Update to error model capable of extracting more appropriate error terms for additive and proportional errors. More error terms are allowed to be converted. ETA conversion to nlmixr has also been reworked, but with the same function.


rikardn (approved):
LGTM!

PR #1805, opened 2023-06-26
Change cov in EstimationStep from bool to string

vrognas:


src/pharmpy/model/external/nonmem/update.py
```diff
@@ -1458,19 +1459,37 @@ def update_estimation(control_stream, model):
...
 
-    old_cov = False
+    # Initiate old_cov
+    old_cov = None
     for est in old:
-        old_cov |= est.cov
-    new_cov = False
+        if est.cov is None:
+            old_cov = None
+        elif est.cov is not None and est.cov in EstimationStep.supported_covs:
```
rikardn, line 1467: No need to check if in supported_covs since this is an invariant. If it is not None it has to be supported.

stellabelin (changes requested):
Looks good! Some minor comments (only the first one is necessary), `docs-test` are also failing.

src/pharmpy/model/external/nonmem/update.py
```diff
@@ -1458,19 +1459,37 @@ def update_estimation(control_stream, model):
...
+            )
+
+    # Initiate new_cov
+    new_cov = None
     for est in new:
-        new_cov |= est.cov
-    if not old_cov and new_cov:
+        if est.cov is None:
+            new_cov = None
+        elif est.cov is not None and est.cov in EstimationStep.supported_covs:
```
rikardn, line 1479: Same as above

rikardn (changes requested):
Thanks for the PR. I found one thing in update

stellabelin (approved):
LGTM!

PR #1814, opened 2023-06-29
Fix string representation in EstimationStep.__repr__()

vrognas:
Ensure strings for method, solver, and cov are correctly enclosed within single quotation marks


stellabelin (approved):
LGTM!

PR #1846, opened 2023-07-10
Add bioavailability statement to model

JohanWBorg:
Possibility of adding bioavailability using a new parameter or simply a number. When transforming the model, bioavailability will move with the dose.

Also fixed to remove lag time when adding transit compartments

### DIFF
--- a/src/pharmpy/model/external/nonmem/update.py
+++ b/src/pharmpy/model/external/nonmem/update.py
@@ -720,17 +720,21 @@ def update_lag_time(model: Model, old: CompartmentalSystem, new: CompartmentalSy
         )
     return model
 
+
 def update_bio(model, old, new):
     new_dosing = new.dosing_compartment[0]
     new_bio = new_dosing.bioavailability
     old_bio = old.dosing_compartment[0].bioavailability
-    
+
     if new_bio != old_bio:
         if isinstance(new_bio, sympy.Symbol):
             if new_bio != sympy.Symbol('F1'):
-                model = model.replace(statements = model.statements.subs({new_bio: sympy.Symbol('F1')}))
-    
+                model = model.replace(
+                    statements=model.statements.subs({new_bio: sympy.Symbol('F1')})
+                )
+
     return model
+
 
 def new_compartmental_map(cs: CompartmentalSystem):
     """Create compartmental map for updated model
--- a/src/pharmpy/model/external/nonmem/update.py
+++ b/src/pharmpy/model/external/nonmem/update.py
@@ -729,17 +729,18 @@ def update_bio(model, old, new):
     """
     if len(new.dosing_compartment) == len(old.dosing_compartment):
         for new_dosing, old_dosing in zip(new.dosing_compartment, old.dosing_compartment):
-            
             new_bio = new_dosing.bioavailability
             old_bio = old_dosing.bioavailability
-            
+
             old_comp_number = model.internals.compartment_map[old_dosing.name]
             newmap = new_compartmental_map(model.statements.ode_system)
             new_comp_number = newmap[new_dosing.name]
             if new_bio != old_bio:
                 if isinstance(new_bio, sympy.Symbol):
                     model = model.replace(
-                        statements=model.statements.subs({new_bio: sympy.Symbol(f'F{new_comp_number}')})
+                        statements=model.statements.subs(
+                            {new_bio: sympy.Symbol(f'F{new_comp_number}')}
+                        )
                     )
             else:
                 if new_bio != sympy.Symbol("F1"):
@@ -747,7 +748,13 @@ def update_bio(model, old, new):
                     # be changed to match the old compartment
                     # All other Fn should be changed to match the new number
                     model = model.replace(
-                        statements=model.statements.subs({sympy.Symbol(f'F{old_comp_number}'): sympy.Symbol(f'F{new_comp_number}')})
+                        statements=model.statements.subs(
+                            {
+                                sympy.Symbol(f'F{old_comp_number}'): sympy.Symbol(
+                                    f'F{new_comp_number}'
+                                )
+                            }
+                        )
                     )
     else:
         # TODO : Number of dose compartments have changed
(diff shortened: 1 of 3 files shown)


src/pharmpy/modeling/odes.py
```diff
@@ -206,6 +206,84 @@ def set_first_order_elimination(model: Model):
...
+        raise ValueError(f'Model {model.name} has no ODE system')
+
+    dose_comp = odes.dosing_compartment[0]
+    bio = dose_comp.bioavailability
+
+    if isinstance(bio, sympy.Number):
+        # Bio not defined
+        if add_parameter:
+            model, bio_symb = _add_parameter(model, 'BIO', init=float(bio))
+            ass = Assignment(sympy.Symbol('F'), bio_symb)
```
stellabelin, line 247: If you use `Assignment.create()` it will automatically make it into a `sympy.Symbol` (same for numbers), see: https://github.com/pharmpy/pharmpy/blob/f357a01c2242be765dc93ddd7b94358d153f88fd/src/pharmpy/model/statements.py#L76-L89

src/pharmpy/modeling/odes.py
```diff
@@ -206,6 +206,84 @@ def set_first_order_elimination(model: Model):
...
+                    + model.statements.before_odes
+                    + f_ass
+                    + CompartmentalSystem(cb)
+                    + model.statements.after_odes
+                )
+            )
+
+    else:
+        # BIO already defined, leave it alone?
+        pass
```
stellabelin, line 282: I guess it makes sense to not do anything, another option would be to warn or to raise, but we don't do that for e.g. `add_lag_time()`.

src/pharmpy/modeling/odes.py
```diff
@@ -765,6 +845,8 @@ def set_transit_compartments(model: Model, n: int, keep_depot: bool = True):
             n -= 1
             cb.add_flow(new_comp, comp, rate)
             comp = new_comp
+        comp = cb.set_bioavailability(comp, dosing_comp.bioavailability)
+        dosing_comp = cb.set_bioavailability(dosing_comp, sympy.Integer(1))
```
stellabelin, line 849: Will this be a problem if you e.g. first have one transit and then add another (i.e. call `set_transit_compartments` multiple times)? Then `dosing_comp` will no longer have the bioavailability?

stellabelin (commented):
Looks good, some comments!

src/pharmpy/model/external/nonmem/update.py
```diff
@@ -720,6 +721,21 @@ def update_lag_time(model: Model, old: CompartmentalSystem, new: CompartmentalSy
     return model
 
 
+def update_bio(model, old, new):
```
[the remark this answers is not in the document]
JohanWBorg: This function has been changed to update all bioavailability compartments instead of only the first one.

src/pharmpy/model/external/nonmem/update.py
```diff
@@ -761,7 +777,8 @@ def pk_param_conversion(model: Model, advan, trans):
     d = {}
     for old, new in remap.items():
         d[sympy.Symbol(f'S{old}')] = sympy.Symbol(f'S{new}')
-        d[sympy.Symbol(f'F{old}')] = sympy.Symbol(f'F{new}')
```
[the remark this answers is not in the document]
JohanWBorg: This is commented out since the bioavailability needs to be moved along with the dose compartment (just as for lag time). If not, then for instance a bioavailability F1 that is moved when adding transit compartment(s) will be renumbered as previous compartment 1 has now become another number (for instance 3 if adding 2 transit comps). This is avoided by instead moving bioavailability with the dose(s)

src/pharmpy/modeling/odes.py
```diff
@@ -206,6 +206,84 @@ def set_first_order_elimination(model: Model):
     return model
 
 
+def add_bioavailability_statement(model: Model, add_parameter: bool = True):
```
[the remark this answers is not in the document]
JohanWBorg: Changed the name in accordance with your suggestion! :)

src/pharmpy/modeling/odes.py
```diff
@@ -206,6 +206,84 @@ def set_first_order_elimination(model: Model):
...
+    """Add bioavailability statement for the dose compartment(s) of the model.
+    Can be added as a new parameter or simply as a number.
```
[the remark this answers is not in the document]
JohanWBorg: Changed the docstring to make clear that the first dose compartment is the one being affected along with specifying that the value is set to 1 if a parameter is not added

src/pharmpy/modeling/odes.py
```diff
@@ -206,6 +206,84 @@ def set_first_order_elimination(model: Model):
...
+
+            model = model.replace(
+                statements=(
+                    bio_ass
+                    + model.statements.before_odes
+                    + f_ass
+                    + CompartmentalSystem(cb)
+                    + model.statements.after_odes
+                )
+            )
```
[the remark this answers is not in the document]
JohanWBorg: Changed the order to be able to remove some rows of code! Nice suggestion!

src/pharmpy/modeling/odes.py
JohanWBorg: It now uses Assignment.create() instead.

src/pharmpy/modeling/odes.py
JohanWBorg: Nothing changed, but kept else statement in case of future updates to the code

src/pharmpy/modeling/odes.py
JohanWBorg: When calling `set_transit_compartments` multiple times the dose compartment is changed for each call and hence the bioavailability is moved for each call. I might be misunderstanding your question but through my own testing this did not cause any issues.
Please let me know if you wish to take a closer look at this.

stellabelin: Yes you're right, misread where in the overall code this was, my mistake!

stellabelin (approved):
LGTM!

PR #1863, opened 2023-07-17
Possibility of parsing IV+ORAL models based on CMT column

JohanWBorg:
Added possibility of parsing nonmem models with doses to mutliple (different) compartments based on the presence of a CMT column. The dosing() function has been changed to extract the doses based on the presence of the CMT column.

Due to pharmpy lacking datasets for testing these sorts of models, testing has mainly been done locally. If a dataset where to be uploaded, i would love to integrate tests for this implementation, or if any other suggestions come up :)

### DIFF
--- a/src/pharmpy/model/external/nonmem/advan.py
+++ b/src/pharmpy/model/external/nonmem/advan.py
@@ -212,7 +212,7 @@ def _compartmental_model(
         cb = CompartmentalSystemBuilder()
         central = Compartment.create(
             'CENTRAL',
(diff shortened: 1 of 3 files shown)


stellabelin (approved):
LGTM!

stellabelin:
Would be good to have a dataset to test the parsing, maybe you could create a test dataset on the fly and write down similarly to what we do in the [integration tests](https://github.com/pharmpy/pharmpy/blob/79e97dce59859d6cb0284a64f6610b60515dbe47/tests/integration/test_allometry.py#L11)?

PR #1868, opened 2023-07-18
Bioavailability update

JohanWBorg:
Provide an update for a simplified way of updating bioavailability regarding Fn numbering. Since bioavailability is moved between compartments during ODE transformations, if the numbering of bioavailability does not match after such a transformation, this update will now renumber this values based on the new ODE structure.

Also added function to remove bioavailability (for the first dose compartment) along with a few unit tests


stellabelin (approved):
LGTM!

Issue #1589, opened 2023-03-20
Support residual error models for multiple dvs


rikardn:
- [x] has_additive_error_model
- [x] has_proportional_error_model
- [x] has_combined_error_model
- [x] additive error
- [x] proportional error
- [x] combined error
- [x] time varying error
- [ ] power
- [x] IIV on RUV

Need to be able to specify which dv has which error model.

JohanWBorg:
As of #1892, the last feature (power) should be implemented

PR #1892, opened 2023-07-27
Support power residual error models for multiple dvs 

JohanWBorg:
Solution to the power model from issue #1589 (adding power model for multiple DVs)
This function can take in either a nothing at all or a specific DV and/or a list of epsilons to perform the transformation on. 

If nothing is specified, the function will perform the transformation on all epsilons that are available, regardless of the affected DV. I was unsure if this should be the case or if only the first (or only) DV should be affected instead, but I decided to follow the same principle as applied to set_iiv_on_ruv().

If a DV is specified, the model will only transform the epsilons connected to said DV (or the ones specified in the list of epsilons). 

If only a list of epsilons is provided, these will all be transformed, regardless of the connected DV.

Please let me know if there are any comments or changes to be made!

### DIFF
--- a/src/pharmpy/modeling/error.py
+++ b/src/pharmpy/modeling/error.py
@@ -3,6 +3,7 @@
 """
 from __future__ import annotations
 
+import warnings
 from typing import List, Optional, Union
 
 from pharmpy.deps import sympy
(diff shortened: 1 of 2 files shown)


src/pharmpy/modeling/error.py
```diff
@@ -895,22 +896,31 @@ def set_time_varying_error_model(
 def set_power_on_ruv(
     model: Model,
     list_of_eps: Optional[Union[str, list]] = None,
+    dv: Union[sympy.Symbol, str, int, None] = None,
     lower_limit: Optional[float] = 0.01,
     ipred: Optional[Union[str, sympy.Symbol]] = None,
     zero_protection: bool = False,
 ):
-    """Applies a power effect to provided epsilons.
+    """Applies a power effect to provided epsilons. If a dependent variable
```
rikardn, line 904: Note that the first line in the docstring is the "short" description of the function. Extended description should come later. I'll merge anyway and this can be fixed later.

rikardn:
Thanks! LGTM (only one minor comment)

PR #1900, opened 2023-07-31
Multiple doses per compartment and parsing updates for IV+ORAL

JohanWBorg:
Added update to both support multiple doses per compartment as  well as the parsing ability of these models. Models with multiple doses require either a CMT column or an admid column in order to work correctly. Changes some naming convention to be more streamline with the number of doses. For instance
ode_system.dosing_compartment -> dosing_compartments
Compartment.dose -> Compartment.first_dose (if only interested in dose, as .dose will return a tuple)

Let me know if there are any comments

### DIFF
--- a/src/pharmpy/model/external/nonmem/advan.py
+++ b/src/pharmpy/model/external/nonmem/advan.py
@@ -305,6 +305,7 @@ def _compartmental_model(
         cs = to_compartmental_system(func_to_name, eqs)
         central_comp_name = cs.central_compartment.name
         cb = CompartmentalSystemBuilder(cs)
+        doses = dosing(di, dataset, 1)
         for i, comp_name in enumerate(comps, start=1):
             comp = cs.find_compartment(comp_name)
             if comp is None:  # Compartments can be in $MODEL but not used in $DES
@@ -315,7 +316,6 @@ def _compartmental_model(
                 central = True
             else:
                 central = False
-            doses = dosing(di, dataset, i)
             cb.set_dose(comp, find_dose(doses, i, admid=admid, central=central))
             comp = cb.find_compartment(comp_name)
             f = _get_bioavailability(control_stream, i)
--- a/src/pharmpy/model/external/nonmem/advan.py
+++ b/src/pharmpy/model/external/nonmem/advan.py
@@ -38,13 +38,18 @@ def _compartmental_model(
     trans,
     des=None,
 ):
+    print("TJA")
+    print(advan)
+    print(advan == "ADVAN1")
     if advan == 'ADVAN1':
         # FIXME : Require multiple doses per comp before IV+ORAL
+        print()
         doses = dosing(di, dataset, 1)
+        print(doses)
         cb = CompartmentalSystemBuilder()
         central = Compartment.create(
             'CENTRAL',
-            dose=find_dose(doses, comp_number=1, admid=2),
+            dose=find_dose(doses, comp_number=1, admid=2, central=True),
             lag_time=_get_alag(control_stream, 1),
             bioavailability=_get_bioavailability(control_stream, 1),
         )
@@ -79,7 +84,7 @@ def _compartmental_model(
         cb = CompartmentalSystemBuilder()
         central = Compartment.create(
             'CENTRAL',
-            dose=find_dose(doses, comp_number=1, admid=2),
+            dose=find_dose(doses, comp_number=1, admid=2, central=True),
             lag_time=_get_alag(control_stream, 1),
             bioavailability=_get_bioavailability(control_stream, 1),
         )
@@ -102,13 +107,13 @@ def _compartmental_model(
         cb = CompartmentalSystemBuilder()
         depot = Compartment.create(
             'DEPOT',
-            dose=find_dose(doses, comp_number=1, admid=1),
+            dose=find_dose(doses, comp_number=1, admid=1, depot=True),
             lag_time=_get_alag(control_stream, 1),
             bioavailability=_get_bioavailability(control_stream, 1),
         )
         central = Compartment.create(
             'CENTRAL',
-            dose=find_dose(doses, comp_number=2, admid=2),
+            dose=find_dose(doses, comp_number=2, admid=2, central=True),
             lag_time=_get_alag(control_stream, 2),
             bioavailability=_get_bioavailability(control_stream, 2),
         )
@@ -196,7 +201,7 @@ def _compartmental_model(
         cb = CompartmentalSystemBuilder()
         central = Compartment.create(
             'CENTRAL',
-            dose=find_dose(doses, comp_number=1, admid=2),
+            dose=find_dose(doses, comp_number=1, admid=2, central=True),
             lag_time=_get_alag(control_stream, 1),
             bioavailability=_get_bioavailability(control_stream, 1),
         )
@@ -212,7 +217,7 @@ def _compartmental_model(
         cb = CompartmentalSystemBuilder()
         central = Compartment.create(
             'CENTRAL',
-            dose=find_dose(doses, comp_number=1, admid=2),
+            dose=find_dose(doses, comp_number=1, admid=2, central=True),
             lag_time=_get_alag(control_stream, 1),
             bioavailability=_get_bioavailability(control_stream, 1),
         )
@@ -244,13 +249,13 @@ def _compartmental_model(
         cb = CompartmentalSystemBuilder()
         depot = Compartment.create(
             'DEPOT',
-            dose=find_dose(doses, comp_number=1, admid=1),
+            dose=find_dose(doses, comp_number=1, admid=1, depot=True),
             lag_time=_get_alag(control_stream, 1),
             bioavailability=_get_bioavailability(control_stream, 1),
         )
         central = Compartment.create(
             'CENTRAL',
-            dose=find_dose(doses, comp_number=2, admid=2),
+            dose=find_dose(doses, comp_number=2, admid=2, central=True),
             lag_time=_get_alag(control_stream, 2),
             bioavailability=_get_bioavailability(control_stream, 2),
         )
@@ -312,8 +317,9 @@ def _compartmental_model(
             admid = 1
             if comp.name == central_comp_name:
                 admid = 2
+                central = True
             doses = dosing(di, dataset, i)
-            cb.set_dose(comp, find_dose(doses, i, admid=admid))
+            cb.set_dose(comp, find_dose(doses, i, admid=admid, central=central))
             comp = cb.find_compartment(comp_name)
             f = _get_bioavailability(control_stream, i)
             cb.set_bioavailability(comp, f)
@@ -628,6 +634,7 @@ def dosing(di: DataInfo, dataset, dose_comp: int):
                  },
             )
     else:
+        print("JORÅ^")
         # CMT column present
         cmt = dataset['CMT']
         if len(cmt.unique()) == 1:
@@ -678,16 +685,25 @@ def dosing(di: DataInfo, dataset, dose_comp: int):
             doses = tuple()
             for comp_number in cmt.unique():
                 cmt_dataset = dataset[dataset['CMT'] == comp_number]
+                print("CMT",comp_number)
                 if 'ADMID' in di.names:
                     # Go through all dose types to the same compartment
                     for admid in cmt_dataset['ADMID'].unique():
+                        if admid == 1:
+                            central = False
+                            depot = True
+                        elif admid == 2:
+                            central = True
+                            depot = False
+                        else:
+                            raise ValueError(f'Administration ID : {admid} not supported')
                         doses += (
                             {
                                 'comp_number': None,
                                 'dose': _dosing(di, cmt_dataset, comp_number),
                                 'admid': admid,
-                                'central': False,
-                                'depot': False,
+                                'central': central,
+                                'depot': depot,
                             },
                         )
                 else:
--- a/src/pharmpy/model/external/nonmem/advan.py
+++ b/src/pharmpy/model/external/nonmem/advan.py
@@ -600,7 +600,9 @@ def dosing(di: DataInfo, dataset, dose_comp: int):
     cmt_loop = False
     admid_loop = False
     return_dose = False
-    if 'ADMID' in di.names:
+    if dataset is None:
+        return_dose = True
+    elif 'ADMID' in di.names:
         for admid in dataset['ADMID'].unique():
             if admid not in [1, 2]:
                 raise ValueError(f'Administration ID : {admid} not supported')
--- a/src/pharmpy/model/external/nonmem/advan.py
+++ b/src/pharmpy/model/external/nonmem/advan.py
@@ -44,7 +44,7 @@ def _compartmental_model(
         cb = CompartmentalSystemBuilder()
         central = Compartment.create(
             'CENTRAL',
-            dose=find_dose(doses, comp_number=1, admid=2, central=True),
+            doses=find_dose(doses, comp_number=1),
             lag_time=_get_alag(control_stream, 1),
             bioavailability=_get_bioavailability(control_stream, 1),
         )
@@ -57,13 +57,13 @@ def _compartmental_model(
         cb = CompartmentalSystemBuilder()
         depot = Compartment.create(
             'DEPOT',
-            dose=find_dose(doses, comp_number=1, admid=1),
+            doses=find_dose(doses, comp_number=1),
             lag_time=_get_alag(control_stream, 1),
             bioavailability=_get_bioavailability(control_stream, 1),
         )
         central = Compartment.create(
             'CENTRAL',
-            dose=find_dose(doses, comp_number=2, admid=2, central=True),
+            doses=find_dose(doses, comp_number=2),
             lag_time=_get_alag(control_stream, 2),
             bioavailability=_get_bioavailability(control_stream, 2),
         )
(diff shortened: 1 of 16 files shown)


src/pharmpy/model/statements.py
```diff
@@ -1416,6 +1557,41 @@ def amount(self):
...
+    @property
+    def number_of_oral_doses(self):
+        return len(self.oral_dose)
+
+    @property
+    def number_of_iv_doses(self):
+        return len(self.iv_dose)
+
+    @property
+    def first_dose(self):
```
stellabelin, line 1592: Indexing can be used instead of a new attribute.

src/pharmpy/model/statements.py
```diff
@@ -1416,6 +1557,41 @@ def amount(self):
...
+
+    @property
+    def iv_dose(self):
+        if self._dose is not None:
+            iv_dose = tuple([dose for dose in self.dose if dose.admid == 2])
+            return iv_dose if iv_dose else None
+        return None
+
+    @property
+    def number_of_doses(self):
```
stellabelin, line 1580: Is this property needed? Using `len()` is not so much longer?

src/pharmpy/model/statements.py
```diff
@@ -1416,6 +1557,41 @@ def amount(self):
...
+            oral_dose = tuple([dose for dose in self.dose if dose.admid == 1])
+            return oral_dose if oral_dose else None
+        return None
+
+    @property
+    def iv_dose(self):
+        if self._dose is not None:
+            iv_dose = tuple([dose for dose in self.dose if dose.admid == 2])
+            return iv_dose if iv_dose else None
+        return None
```
stellabelin, line 1577: Are these used anywhere? Maybe it's better to wait to add these methods until they're needed?

src/pharmpy/modeling/odes.py
```diff
@@ -206,7 +208,9 @@ def set_first_order_elimination(model: Model):
     return model
 
 
-def add_bioavailability(model: Model, add_parameter: bool = True):
+def add_bioavailability(
```
stellabelin, line 211: Do you need the option here to set the initial estimate since we have functions for that? Was thinking so that it's equivalent to `add_lag_time()`, and also the initial estimate is only relevant if you add a parameter

src/pharmpy/modeling/odes.py
```diff
@@ -237,13 +243,15 @@ def add_bioavailability(model: Model, add_parameter: bool = True):
...
-    dose_comp = odes.dosing_compartment[0]
+    dose_comp = odes.first_dosing_compartment
     bio = dose_comp.bioavailability
 
     if isinstance(bio, sympy.Number):
         # Bio not defined
         if add_parameter:
-            model, bio_symb = _add_parameter(model, 'BIO', init=float(bio))
+            model, bio_symb = _add_parameter(
+                model, 'BIO', init=float(init), lower=float(lower), upper=float(upper)
```
stellabelin, line 253: No need to explicitly convert to `float` I think, in `add_population_parameter` it will use the `Parameter.create()` method.

stellabelin (changes requested):
Looks good, but have some comments! In general, several are related to whether a new method/property is really needed or if current ones can be used.

src/pharmpy/model/statements.py
```diff
@@ -367,6 +389,94 @@ def set_dose(self, compartment, dose):
         nx.relabel_nodes(self._g, mapping, copy=False)
         return new_comp
 
+    def replace_dose(self, compartment, old_dose, new_dose):
```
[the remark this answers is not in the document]
rikardn: I agree that this would make more sense

src/pharmpy/model/statements.py
```diff
@@ -940,6 +1050,28 @@ def dosing_compartment(self):
 
         raise ValueError('No dosing compartment exists')
 
+    @property
+    def first_dosing_compartment(self):
```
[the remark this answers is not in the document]
rikardn: Agree `odes.dosing_compartments[0]` should be enough

src/pharmpy/model/statements.py
```diff
@@ -1374,8 +1509,14 @@ def create(
             amount = parse_expr(amount)
         else:
             amount = sympy.Symbol(f'A_{name}')
-        if dose is not None and not isinstance(dose, Dose):
-            raise TypeError("dose must be of Dose type (or None)")
+        if isinstance(dose, Dose):
+            dose = (dose,)
+        if dose is not None and not isinstance(dose, tuple):
```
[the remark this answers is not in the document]
rikardn: So no dose would be an empty `tuple`? I think that makes sense. The `None` case is then not needed. `bool` for empty tuple should be `False`

src/pharmpy/model/statements.py
rikardn: Do we even have a way of distinguishing the oral dose from the iv?

rikardn (commented):
I agree with Stellas comments. Since there are many changes/updates to the core API I think we should both review again after fixes have been applied.

src/pharmpy/model/statements.py
```diff
@@ -347,8 +347,30 @@ def move_dose(self, source, destination):
         mapping = {source: new_source, destination: new_dest}
         nx.relabel_nodes(self._g, mapping, copy=False)
 
+    def move_oral_dose(self, source, destination, single_dose=True):
```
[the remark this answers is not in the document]
JohanWBorg: Yes, `move_dose` has been updated with and argument `admid` instead of adding a new method. Also added argument `replace` to determine of the moved dose should replace the dose of the destination compartment or not

src/pharmpy/model/statements.py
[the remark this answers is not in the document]
JohanWBorg: That was precisely the function. To avoid adding another method, your suggestion has been implemented, replacing based on `admid` as an option in `set_dose`

src/pharmpy/model/external/nonmem/advan.py
```diff
@@ -590,36 +593,102 @@ def _advan12_trans(trans: str):
...
+            # All doses to same compartment
+            if 'ADMID' in di.names:
+                warnings.warn(
+                    "CMT column present with only one value" "Using ADMID to determine dose type"
+                )
+                # Go through all dose types to the same compartment
+                for admid in dataset['ADMID'].unique():
+                    if admid not in [1, 2]:
+                        raise ValueError(f'Administration ID : {admid} not supported')
+                    doses += (
```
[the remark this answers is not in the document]
JohanWBorg: It had not and went under the radar. But as of the reformat (see next comment) this is not an issue any more

src/pharmpy/model/external/nonmem/advan.py
```diff
@@ -590,36 +593,102 @@ def _advan12_trans(trans: str):
 
 
 def dosing(di: DataInfo, dataset, dose_comp: int):
+    # Only check doses
+    if dataset is not None:
+        dataset = dataset[dataset['AMT'] != 0]
+
     if 'CMT' not in di.names or di['CMT'].drop or dataset is None:
-        return ({'comp_number': dose_comp, 'dose': _dosing(di, dataset, dose_comp)},)
+        if 'ADMID' in di.names:
```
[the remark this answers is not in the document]
JohanWBorg: Yes, that is about the functionality of the function. But instead of only having different subsets, the function will loop through different parts of the dataset (such as the CMT column or the admid column). I have reformatted the structure for readability but please let me know if there are any issues :)

src/pharmpy/model/external/nonmem/advan.py
```diff
@@ -633,12 +702,24 @@ def _dosing(di, dataset, dose_comp, cmt=False):
...
+                and not dose['admid'] == 2
+            )
+        ):
             comp_dose = dose['dose']
-            return comp_dose.replace(admid=admid)
-    return None
+            if dose['admid'] is not None:
+                admid = dose['admid']
+            comp_doses += (comp_dose.replace(admid=admid),)
+    return comp_doses if comp_doses else None
```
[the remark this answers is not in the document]
JohanWBorg: This has been changed to return an empty tuple

src/pharmpy/model/statements.py
```diff
@@ -367,6 +389,94 @@ def set_dose(self, compartment, dose):
...
+        if compartment.dose is None:
+            new_comp = compartment.replace(dose=(dose,))
+        else:
+            new_dose = compartment.dose + (dose,)
+            new_comp = compartment.replace(dose=tuple(sorted(new_dose, key=lambda d: d.admid)))
+        mapping = {compartment: new_comp}
+        nx.relabel_nodes(self._g, mapping, copy=False)
+        return new_comp
+
+    def remove_dose(self, compartment, dose_to_remove):
```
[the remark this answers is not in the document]
JohanWBorg: This was a special case for how `set_first_dose_absorption` was previously implemented. But I have worked around this in order to remove `remove_dose` as, just as you mention, in should perhaps not be needed.

src/pharmpy/model/statements.py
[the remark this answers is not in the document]
JohanWBorg: The None case has been removed and `.dose` for compartments with no dose is now an empty tuple

src/pharmpy/model/statements.py
```diff
@@ -1416,6 +1557,41 @@ def amount(self):
...
+    def number_of_doses(self):
+        return len(self._dose)
+
+    @property
+    def number_of_oral_doses(self):
+        return len(self.oral_dose)
+
+    @property
+    def number_of_iv_doses(self):
+        return len(self.iv_dose)
```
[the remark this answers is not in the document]
JohanWBorg: This was done based on the connected admid but have now been removed.

src/pharmpy/model/statements.py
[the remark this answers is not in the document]
JohanWBorg: Changed back to indexing.

src/pharmpy/model/statements.py
JohanWBorg: Also changed back to indexing in all cases.

src/pharmpy/model/statements.py
JohanWBorg: No, and has now been removed to avoid adding unnecessary properties

src/pharmpy/model/statements.py
JohanWBorg: This was done base on admid and where used in some special places for ODE transformations. However, as they are not necessarily needed, they have been removed for now.

src/pharmpy/modeling/odes.py
JohanWBorg: In order to follow the same logic as for `add_lag_time`, that option has been removed in order to stay consistent.

JohanWBorg:
These changes address the concerns you mentioned in your comments. As stated, most are for deletions of properties/functions in order to avoid bloating and implement the changes within existing functions instead.
Please let me know if any other changes are needed or if there are any other questions :)

src/pharmpy/modeling/odes.py
stellabelin: Ah okay I was a bit hasty, it seems we do this conversion for `upper` and `lower`, but not `init`:

https://github.com/pharmpy/pharmpy/blob/87bfb030b0456978061624d8cfabf5840891d4e6/src/pharmpy/model/parameters.py#L67-L74

Would it make sense to do so here as well? Or is there a particular reason it's not done for `init` @rikardn? Was thinking it would be nice to not have to think about it when using the `Parameter.create()` method.

src/pharmpy/model/external/nonmem/advan.py
```diff
@@ -590,36 +595,92 @@ def _advan12_trans(trans: str):
...
+    # Only check doses
+    if dataset is not None:
+        dataset = dataset[dataset['AMT'] != 0]
+
+    cmt_loop = False
+    admid_loop = False
+    return_dose = False
+    if dataset is None:
+        return_dose = True
+    elif 'ADMID' in di.names:
```
rikardn, line 607: The name ADMID shouldn't be assumed. Look for the 'admid' type using the type indexer. ('CMT' is safe to assume since it is a reserved name)

src/pharmpy/model/external/nonmem/update.py
```diff
@@ -441,11 +442,11 @@ def update_cmt_column(model, old, new):
...
             remap = create_compartment_remap(oldmap, newmap)
 
-            for dose_comp in old.dosing_compartment:
+            for dose_comp in old.dosing_compartments:
                 if dose_comp != old.central_compartment:
                     # Remap oral doses to new dosing compartment
-                    remap[oldmap[dose_comp.name]] = newmap[new.dosing_compartment[0].name]
-
+                    remap[oldmap[dose_comp.name]] = newmap[new.dosing_compartments[0].name]
+            dataset = model.dataset.copy()
```
rikardn, line 449: No need to copy here. The `replace` method has `inplace=False`

src/pharmpy/model/statements.py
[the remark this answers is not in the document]
rikardn: If you don't use it you should remove it

src/pharmpy/model/statements.py
```diff
@@ -649,7 +712,7 @@ def subs(self, substitutions):
         >>> from pharmpy.modeling import load_example_model
         >>> model = load_example_model("pheno")
         >>> model.statements.ode_system.subs({'AMT': 'DOSE'})
-        Bolus(DOSE, admid=1)
+        Bolus(DOSE, admid=2) --> CENTRAL
```
rikardn, line 715: This looks strange. Where does the  `--> CENTRAL` come from?

src/pharmpy/model/statements.py
```diff
@@ -814,7 +877,7 @@ def get_compartment_inflows(self, compartment):
         >>> from pharmpy.modeling import load_example_model
         >>> model = load_example_model("pheno")
         >>> model.statements.ode_system.get_compartment_inflows(output)
-        [(Compartment(CENTRAL, amount=A_CENTRAL, dose=Bolus(AMT, admid=1)), CL/V)]
+        [(Compartment(CENTRAL, amount=A_CENTRAL, dose=Bolus(AMT, admid=2)), CL/V)]
```
rikardn, line 880: Why do we have `admid=2` here? Only one route should mean `admid=1` as default?

src/pharmpy/model/statements.py
```diff
@@ -1318,7 +1384,7 @@ class Compartment:
     ----------
     name : str
         Compartment name
-    dose : Dose
+    dose : Dose, tuple(Dose)
```
rikardn, line 1387: Don't allow `Dose` here. The standard constructor should only take the type of the attribute.

src/pharmpy/model/statements.py
```diff
@@ -1374,8 +1440,16 @@ def create(
             amount = parse_expr(amount)
         else:
             amount = sympy.Symbol(f'A_{name}')
-        if dose is not None and not isinstance(dose, Dose):
-            raise TypeError("dose must be of Dose type (or None)")
+        if dose is None:
+            dose = tuple()
+        if isinstance(dose, Dose):
+            dose = (dose,)
+        if not isinstance(dose, tuple):
```
rikardn, line 1447: Convert to tuple instead so that any sequence can be supported.

src/pharmpy/model/statements.py
```diff
@@ -1374,8 +1440,16 @@ def create(
...
+            raise TypeError("dose(s) need to be given as a tuple")
+        if dose is not None:
```
rikardn, line 1449: Not needed it cannot be `None` here

src/pharmpy/model/statements.py
```diff
@@ -1416,7 +1490,10 @@ def amount(self):
 
     @property
     def dose(self):
-        return self._dose
+        if self._dose is not None:
```
rikardn, line 1493: This cannot happen. `_dose` must be a `tuple`

rikardn (changes requested):

src/pharmpy/model/statements.py
JohanWBorg: `admid=1` is defined to represent an oral dose and `amid=2` represents an IV dose. Since this is a one-compartment system with a Bolus dose directly into the CENTRAL compartment, I believe `admid=2` should be the correct numbering. Or am I missing something?

rikardn: Yes, but only for multiple doses. If there is only one dose it should be `admid=1`. It is only an id and we should really have a way of defining which is which, but until we have that in place I think this makes the most sense.

JohanWBorg: ok, i'll correct this to only change the admid in the presence of multiple doses

src/pharmpy/model/statements.py
```diff
@@ -340,15 +340,34 @@ def move_dose(self, source, destination):
...
+        if replace:
+            new_dest = destination.replace(doses=source.doses)
+        else:
+            new_dest = destination.replace(doses=destination.doses + new_dest_dose)
         mapping = {source: new_source, destination: new_dest}
         nx.relabel_nodes(self._g, mapping, copy=False)
+        return new_source, new_dest
+
+    def set_dose(self, compartment, dose, replace=True, admid=None):
+        """Set dose of compartment, replacing the previous if existing
```
rikardn, line 365: The first line of the docstring is misleading given that replacing is optional

rikardn (approved):
Only some very minor things remain. Can be fixed separately if you want. Main is currently failing so I'll wait with merging.

src/pharmpy/model/external/nonmem/advan.py
```diff
@@ -169,17 +169,15 @@ def _compartmental_model(
...
+        for i, name in enumerate(comp_names, start=1):
+            curdose = find_dose(doses, i)
             comp = Compartment.create(
                 name,
-                dose=curdose,
+                doses=curdose,
                 lag_time=_get_alag(control_stream, i),
                 bioavailability=_get_bioavailability(control_stream, i),
             )
+            print(comp)
```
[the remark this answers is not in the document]
JohanWBorg: Great catch :D Must have been missed while going over the changed files, thanks!

src/pharmpy/model/statements.py
JohanWBorg: Changed to "Set dose of compartment, replacing the previous _by default_". Also added the `replace` and `admid` to "Parameters"

PR #1943, opened 2023-08-10
Removed ValueError if trying to add power to non-matching IPRED symbol

JohanWBorg:
Allow the user to add power model to a symbol of their choosing.


rikardn (approved):
Looks good.

PR #1944, opened 2023-08-10
Drug metabolite model support + structsearch

JohanWBorg:
Added features for drug metabolite models, including:

- add_metabolite() can add plain metabolite model, with 100 metabolized drug from central to a "METABOLITE" compartment
- add_metabolite() now has option "presystemic_fixed_drug" which, when set to True, will add the "METABOLITE" compartment as presystemic instead. This will also add a DEPOT compartment if needed. "plain" metabolite models can also be converted to presystemic using this function.
- add_peripheral_compartment() has been given an option "name" which can add the peripheral compartment to the compartment with the specified name. The same has been done to remove_peripheral_compartment(). The same has been applied to the property Compartmental.system.peripheral.compartments which has hence been changed to a function.

Structsearch has been developed for drug metabolite models as well with the specified type "drug_metabolite" which will examine four models
- "plain" metabolite w/wo peripheral compartment
- presystemic fixed drug w/wo peripheral compartment


src/pharmpy/model/statements.py
```diff
@@ -1113,7 +1130,10 @@ def _find_depot(self):
         depot = None
         for to_central, _ in self.get_compartment_inflows(central):
             outflows = self.get_compartment_outflows(to_central)
-            if len(outflows) == 1:
+            if len(outflows) == 1 or len(outflows) == 2:
+                if len(outflows) == 2:
+                    if not self.get_flow(to_central, self.find_metabolite()):
```
rikardn, line 1135: Why does the logi for finding the depot need to be changed?

rikardn (changes requested):
I have some comments.

As a general comment/tip it would be good if one commit would add one feature. This not only makes it easier for the reviewer to go through one commit at a time, but it also makes it easier to follow the history and to revert specific features.  One way is to develop from the bottom. Add needed functions/methods first in separate commits and use them in other commits. Use `amend` if formatting was forgotten.

src/pharmpy/model/statements.py
JohanWBorg: This is due to presystemic metabolite models where the depot then will have two outflows (as opposed to one). One going to the central compartment and one going to the metabolite compartment.
As it was previously defined as a compartment having a single outflow to the central compartment, this would be incorrect for presystemic metabolite models

src/pharmpy/model/statements.py
```diff
@@ -1123,6 +1143,19 @@ def _find_depot(self):
                     break
         return depot
 
+    def find_metabolite(self):
```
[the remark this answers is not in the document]
JohanWBorg: Will remove the method

src/pharmpy/model/external/nonmem/update.py
```diff
@@ -795,11 +795,25 @@ def update_bio(model, old, new):
             not isinstance(dose.bioavailability, sympy.Number)
             and dose.bioavailability != f'F{newmap[dose.name]}'
         ):
-            model = model.replace(
-                statements=model.statements.subs(
-                    {sympy.Symbol(f'{dose.bioavailability}'): sympy.Symbol(f'F{newmap[dose.name]}')}
+            if re.match("F[0-9]", str(dose.bioavailability)):
```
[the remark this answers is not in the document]
JohanWBorg: I believe you are right. It seems to have been transferred been duplicated between this PR and PR #1999. I believe it is not required in this PR and will remove it from here.

JohanWBorg:
Thank you for your comments and tips. I will incorporate it in my future work :)

src/pharmpy/model/external/nonmem/update.py
```diff
@@ -1055,6 +1055,8 @@ def match_advan4(odes):
     if len(outflows) != 1:
         return False
     central = outflows[0][0]
+    if central.name == "METABOLITE":
```
rikardn, line 1058: There is a solution to this without using the compartment names in `match_advan2`. Perhaps this can be made into a more general function. For context: the problem is that we are matching an advan that could be used, but the reserved symbols in NONMEM are different so we would rather use the general linear advan.

src/pharmpy/model/statements.py
```diff
@@ -996,19 +998,26 @@ def central_compartment(self):
...
+                central = self.find_compartment("CENTRAL")
+                if central is None:
+                    raise ValueError('Cannot find central compartment')
         except IndexError:
             raise ValueError('Cannot find central compartment')
         return central
 
-    @property
-    def peripheral_compartments(self):
+    def peripheral_compartments(self, name=None):
```
rikardn, line 1012: Since this is no longer a property it should change to a name with a verb. We have a similar situation for transits I think using `find_transit_compartments`. I like using name as the argument here.

src/pharmpy/model/statements.py
```diff
@@ -1113,7 +1130,10 @@ def _find_depot(self):
         depot = None
         for to_central, _ in self.get_compartment_inflows(central):
             outflows = self.get_compartment_outflows(to_central)
-            if len(outflows) == 1:
+            if len(outflows) == 1 or len(outflows) == 2:
+                if len(outflows) == 2:
+                    if not self.get_flow(to_central, self.find_compartment("METABOLITE")):
```
rikardn, line 1135: Is this needed? How do we find the central compartment in the presystemic case?

src/pharmpy/modeling/metabolite.py
```diff
@@ -12,22 +12,28 @@
...
 
 from .error import set_proportional_error_model
-from .odes import _find_noncov_theta, add_individual_parameter
-from .parameters import set_initial_estimates
+from .odes import _find_noncov_theta, add_individual_parameter, set_first_order_absorption
+from .parameters import set_initial_estimates, set_lower_bounds, set_upper_bounds
 
 
-def add_metabolite(model: Model, drug_dvid: int = 1):
+def add_metabolite(model: Model, drug_dvid: int = 1, presystemic_fixed_drug: bool = False):
```
rikardn, line 19: Is presystemic fixed drug the correct term?

rikardn (changes requested):
I have some comments.

src/pharmpy/model/external/nonmem/update.py
JohanWBorg: This has been implemented in the latest changes

src/pharmpy/model/statements.py
JohanWBorg: Changed the naming to find_peripheral_compartments() to be in line with the transits compartments

src/pharmpy/model/statements.py
JohanWBorg: Currently, we have no precise way of differentiating the central and metabolite compartments in the presystemic case. So this is currently dependent on the compartment name as previously discussed.

src/pharmpy/modeling/metabolite.py
JohanWBorg: Changed to "presystemic" as it's a bit simpler more in line with what was found in theory

rikardn:
Thanks!

PR #1966, opened 2023-08-15
Update help_functions.py

davsjob:
Changed `_get_eta_symbs` to prevent a bug accidentally returning EPS. Changing `rvs.free_symbols` to `rvs.etas.free_symbols` in row 56 fixes the bug.


rikardn:
Good catch!

PR #1981, opened 2023-08-23
Remove `LALR(1)` project from `projects.rst`

king-of-poppk:
This was implemented in https://github.com/pharmpy/pharmpy/pull/1304 and https://github.com/pharmpy/pharmpy/commit/66913eafcbf89f6568852c9fbaf155446bd86da5.


rikardn:
Thanks! Good catch!

PR #1984, opened 2023-08-23
Correct nlmixr execute_model() to follow same format as respective NONMEM function

JohanWBorg:
Bug fix for running nlmixr using fit(model, tool="nlmixr")


rikardn:
LGTM

PR #2006, opened 2023-09-01
Refactor naming in EstimationStep class

vrognas:
This PR includes two major changes in the EstimationStep class and elsewhere:

Renaming `cov` to `uncert_method`.
Renaming `method` to `est_method`.

### Why These Changes Are Important:

- Clarity: Renaming `cov` to `uncert_method` distinguishes it from the term "covariate", and the term `est_method` specifies that it's the method for parameter estimation — limiting confusion.
- Consistency: These changes align well with each other, bringing a unified naming convention to the `EstimationStep` class.
- Searchability: The unique terms allows for more targeted code searches.

Please review these changes and let me know if you have any questions, comments, need further clarifications, or have feedback.

### DIFF
--- a/docs/estmethod.rst
+++ b/docs/estmethod.rst
@@ -24,17 +24,17 @@ To initiate Estmethod in Python/R:
     res = run_estmethod(algorithm='exhaustive',
                         model=start_model,
                         results=start_model_results,
-                        methods='all',
+                        est_methods='all',
                         solvers=['LSODA', 'LSODI'])
 
 This will take an input model ``start_model``. The tool will use the 'exhaustive' ``algorithm`` and try all combinations
-between all available ``methods`` and ``solvers`` LSODA and LSODI.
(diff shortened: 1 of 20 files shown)


vrognas:
I wonder if it might be even more beneficial to take it a step further. What if we carved out the uncertainty aspect into its very own class, maybe named `UncertaintyStep`? This way, we could retain the simplicity of calling the arguments `method`. I remember having a chat with @stellabelin where she hinted that the parameter uncertainty is tightly linked with the parameter estimation in the current code structure.

However, I can't shake off the feeling that establishing a separate `UncertaintyStep` class could pave a smoother road for us in the future, especially when it comes to expanding or modifying features. It might offer a cleaner separation and perhaps even make the code more modular and adaptable to changes.

I'd love to hear your thoughts on this and whether you think this could be a viable path to explore. Always open to further discussions or clarifications!

Thanks!

PR #2020, opened 2023-09-07
Add explanatory comments to update_estimation function

vrognas:
This commit adds detailed comments to the update_estimation function to explain the role and possible values of the "op" variable. This enhancement aims to improve code readability and maintainability.


rikardn:
I saw that the `diff` function wasn't imported directly from its module, which makes it hard to find its documentation. Changed that here 38d4c27c600b5794c032a892afa5c3f63fb9c03b. This should make it easier to find the documentation of what comes out of the diff function.  I hope that this will remove the need for the comments.

PR #2033, opened 2023-09-13
Automatically reset model to proportional if needed in RUVsearch

JohanWBorg:
When running RUVsearch, the input model will be checked to see if it has a proportional error model. If it does, then the RUVsearch will move on just as before. If it does not however, a proportional error model will be set and the the model will be re-fitted (in order to obtain new CWRES values) and then the RUVsearch carry on as before (with the proportional error model as input model)
As a final step, the input model is compared to the best resulting model from the search to make sure that an improvement has been made. This effects the cases where the input model was reset to a proportional error model.


rikardn:
LGTM!

PR #2040, opened 2023-09-14
Rename 'cov(s)' to 'parameter_uncertainty_method(s)'

vrognas:
Hello team! 🌟

In this PR, I've made an update to clarify the code. Here's a brief overview:

1. Renaming `cov`: The argument `cov` has been replaced with `uncert_method` throughout various files. This change emphasizes that the argument deals with parameter uncertainty, avoiding any potential confusion with the term "covariates".
2. Avoiding Confusion: The new terminology separates the concept of parameter uncertainty from "covariates", making the code more straightforward and intuitive.

Let me know your thoughts!


rikardn:
I think we should use `parameter_uncertainty_method` instead to make sure that there are no misunderstandings to what we mean.

stellabelin (commented):
I think the code looks good, but I also think `uncert_method` is not super clear, `uncertainty_method` could maybe also work? And then document that it refers to parameter uncertainty?

rikardn:
Ok, I think  `uncertainty_method` is ok.

vrognas:
I agree `uncert_method` is not super precise, even though it is an improvement from `cov`.  `parameter_uncertainty_method` is clear, even if it is a bit long. My preference is for clarity over parsimony, so I am preparing a PR with the name changed to `parameter_uncertainty_method`.

src/pharmpy/tools/estmethod/tool.py
```diff
@@ -26,46 +26,47 @@ def create_workflow(
...
-    results : ModelfitResults
-        Results for model
-    model : Model
-        Pharmpy model
+         List of estimation methods to test.
+         Can be specified as 'all', a list of estimation methods, or None (to not test any estimation method)
+     solvers : list, str or None
+         List of solver to test. Can be specified as 'all', a list of solvers, or None (to
+         not test any solver)
+     parameter_uncertainty_method : list, str or None
```
stellabelin, line 45: Missed this in the last review but it should be `parameter_uncertainty_methods` right?

stellabelin (changes requested):
Just noticed a small spelling error in the docstring for `run_estmethod` that might cause issues on the pharmr side. Otherwise LGTM!

src/pharmpy/tools/estmethod/tool.py
vrognas: Good catch! 🎣

stellabelin (approved):
LGTM!

PR #2041, opened 2023-09-14
Implement drug metabolite within AMD

JohanWBorg:
Possibility of specifying "drug_metabolite" modeltype within the AMD tool which will run structsearch for drug_metabolite.
Also updated structsearch for drug_metabolite to always return a drug metabolite model by using a "plain" type model as base.

In addition to these implementations. Functionality for extracting CL, V and Q parameters connected to a metabolite has been introduced for accurate extraction of all parameters of interest when running drug metabolite models.

### DIFF
--- a/src/pharmpy/tools/structsearch/drugmetabolite.py
+++ b/src/pharmpy/tools/structsearch/drugmetabolite.py
@@ -4,6 +4,28 @@ from pharmpy.model import Model
 from pharmpy.modeling import add_metabolite, add_peripheral_compartment, set_name
 
 
+def create_base_metabolite(model: Model) -> Model:
+    """
+    Create a plain metabolite models from an input model. This is used
+    as the base model when searching for candidate drug metabolite model
+
+    Parameters
+    ----------
+    model : Model
+        Input model for search
+
+    Returns
+    -------
+    Model
+        Base model for drug metabolite model search
+
+    """
+    model = add_metabolite(model)
+    model = set_name(model, "base_metabolite")
+    model = model.replace(parent_model="base_metabolite")
+    return model
+
+
 def create_drug_metabolite_models(model: Model) -> List[Model]:
     """
     Create candidate models for drug metabolite model structsearch.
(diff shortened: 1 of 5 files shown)


src/pharmpy/tools/structsearch/drugmetabolite.py
```diff
@@ -4,6 +4,28 @@
 from pharmpy.modeling import add_metabolite, add_peripheral_compartment, set_name
 
 
+def create_base_drug_metaboltie(model: Model) -> Model:
```
rikardn, line 7: misspelled function name

src/pharmpy/tools/structsearch/drugmetabolite.py
```diff
@@ -4,6 +4,28 @@
...
+        Input model for search
+
+    Returns
+    -------
+    Model
+        Base model for drug metabolite model search
+
+    """
+    model = add_metabolite(model)
+    model = set_name(model, "plain")
```
rikardn, line 24: Perhaps name "base" to match other tools or alternatively "base-metabolite"

src/pharmpy/tools/structsearch/tool.py
```diff
@@ -83,6 +83,7 @@ def run_tmdd(context, model):
 
     wf = create_fit_workflow(qss_candidate_models)
     task_results = Task('results', bundle_results)
+    # FIXME : wf (Workflow) has no attribute called add_task
```
rikardn, line 86: This is correct and needs to be fixed. I won't work to use `add_task` without going via a `WorkflowBuilder`

rikardn: But since it is not related to this PR so no need to fix here.

rikardn (changes requested):
Some comments

src/pharmpy/tools/structsearch/drugmetabolite.py
JohanWBorg: Oh how embarrassing! I corrected the error.

src/pharmpy/tools/structsearch/drugmetabolite.py
JohanWBorg: named "base_metabolite" for clarity of origin

src/pharmpy/tools/structsearch/tool.py
JohanWBorg: Will leave the comment as is for future fix then

rikardn:
Thanks!

PR #2058, opened 2023-09-19
Update $COVARIANCE command with new default options

vrognas:
This commit enriches the $COVARIANCE command with more detailed default options (UNCONDITIONAL, PRINT=E, PRECOND=1), enhancing the output and success rate of the covariance step.

* Modified docstring in estimation.py to explain the new default options.
* Updated update.py to incorporate the new default options in the $COVARIANCE command.
* Adjusted test cases in test_estimation_steps.py and test_nonmem_model.py to reflect and test the new defaults.

These improvements aim to facilitate more insightful and stable model development.


rikardn:
Good idea. I might have excluded `PRINT=E` since it can be calculated if needed, but fair enough

PR #2062, opened 2023-09-20
Auto filter dataset before fitting if missing results for drug_metabolite

JohanWBorg:
If model is missing results when running "drug_metabolite" in AMD, the dataset need to be filtered for DVID=1 before fitting. This filtering is reset during structsearch (after modelsearch has been run)

### DIFF
--- a/src/pharmpy/tools/amd/run.py
+++ b/src/pharmpy/tools/amd/run.py
@@ -270,6 +270,12 @@ def run_amd(
                 f"Unrecognized section {section} in order. Must be one of {default_order}"
             )
 
+    # Filter data to only contain dvid=1
+    if modeltype == "drug_metabolite":
+        orig_dataset = model.dataset
+        # FIXME : remove alongside create_pk_model
+        model = create_pk_model(model)
+
     if results is None:
         model = run_tool('modelfit', model, path=db.path / 'modelfit', resume=resume)
     else:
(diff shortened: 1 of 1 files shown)


src/pharmpy/tools/amd/run.py
```diff
@@ -270,6 +270,12 @@ def run_amd(
                 f"Unrecognized section {section} in order. Must be one of {default_order}"
             )
 
+    # Filter data to only contain dvid=1
```
rikardn, line 273: I don't understand this comment. Where does the filtering happen?

JohanWBorg: In the if-statement below the comment. The original dataset is saved and the model with the filtered dataset is outputted from `create_pk_model`. I believe @antoniaih is working on renaming this function.

rikardn: Ok so this is simply keeping the original dataset to be able to put it back later.

rikardn: If all underlying functions are behaving there should be no need for the defensive copy.

JohanWBorg: Yes precisely. If you are referring to `.copy()` I can remove it. It was mainly added as a safety :)

rikardn:
Ok, now it LGTM!

PR #2066, opened 2023-09-20
Rename arguments in `estmethod`

vrognas:


stellabelin (commented):
The code changes LGTM, however I think that having emojis in commit messages might not be the best since it can be a bit confusing/open to interpretation?

Issue #1899, opened 2023-07-31
Bug in creating models in estmethod when testing with cov option


keqiss:
### Bug report

`res = run_estmethod(algorithm = 'exhaustive',model=model, results=results, covs='all')`

1. The IMP evaluation was added in every model and couldn't be disabled, which makes IMP covariance always output. 
2. The original $EST was overwritten by auto setting e.g. `$ESTIMATION METHOD=COND INTER MAXEVAL=9999 AUTO=1 PRINT=10`, making it impossible to apply $COV with the initial $EST. 

For instance:
```
$ESTIMATION METHOD=IMP INTER MAXEVAL=9999 AUTO=1 PRINT=10
$ESTIMATION METHOD=IMP INTER EONLY=1 MAXEVAL=9999 ISAMPLE=10000 NITER=10 PRINT=10
$COVARIANCE MATRIX=S
```

Pharmpy version: 0.98.2

vrognas:
This issue is really two separate issues. I'll address the first point first.

stellabelin:
There is now an option `compare_ofv` which was added in 01d60ac, setting this to `False` will not add an IMP-evaluation step. Closing the issue.

PR #2068, opened 2023-09-20
Don't always add `IMP EONLY=1` when running `estmethod`

vrognas:
With the modifications in this PR, `IMP EONLY=1` will only be added when multiple estimation methods are to be compared or when the `exhaustive_only_eval` is used.

Previously, the IMP evaluation was automatically added to every model during the run of the `estmethod` tool when only evaluating different `$COV`, and it couldn't be disabled.

Resolves: First point in issue #1899

### DIFF
--- a/src/pharmpy/tools/estmethod/algorithms.py
+++ b/src/pharmpy/tools/estmethod/algorithms.py
@@ -12,12 +12,19 @@ def exhaustive(methods, solvers, parameter_uncertainty_methods):
     task_start = Task('start', start)
     wb.add_task(task_start)
 
+    add_eval_after_est = len(methods) > 1
+
     candidate_no = 1
     for method, solver, parameter_uncertainty_method in itertools.product(
         methods, solvers, parameter_uncertainty_methods
     ):
         wf_estmethod = _create_candidate_model_wf(
-            candidate_no, method, solver, parameter_uncertainty_method, update_inits=False
+            candidate_no,
+            method,
+            solver,
+            parameter_uncertainty_method,
+            add_eval_after_est,
+            update_inits=False,
         )
         wb.insert_workflow(wf_estmethod, predecessors=task_start)
         candidate_no += 1
@@ -28,7 +35,18 @@ def exhaustive(methods, solvers, parameter_uncertainty_methods):
 def exhaustive_with_update(methods, solvers, parameter_uncertainty_methods):
     wb = WorkflowBuilder()
 
-    task_base_model = Task('create_base_model', _create_base_model)
+    add_eval_after_est = len(methods) > 1
+
+    parameter_uncertainty_method_base = (
+        parameter_uncertainty_methods[0] if 'FOCE' in methods else None
+    )
+
+    task_base_model = Task(
+        'create_base_model',
+        _create_base_model,
+        parameter_uncertainty_method_base,
+        add_eval_after_est,
+    )
     wb.add_task(task_base_model)
     wf_fit = create_fit_workflow(n=1)
     wb.insert_workflow(wf_fit, predecessors=task_base_model)
@@ -39,17 +57,31 @@ def exhaustive_with_update(methods, solvers, parameter_uncertainty_methods):
         methods, solvers, parameter_uncertainty_methods
     ):
         # This is equivalent to the base model
-        if not (method == 'FOCE' and solver is None):
+        if not (
+            method == 'FOCE'
+            and solver is None
+            and parameter_uncertainty_method == parameter_uncertainty_methods[0]
+        ):
             # Create model with original estimates
             wf_estmethod_original = _create_candidate_model_wf(
-                candidate_no, method, solver, parameter_uncertainty_method, update_inits=False
+                candidate_no,
+                method,
+                solver,
+                parameter_uncertainty_method,
+                add_eval_after_est,
+                update_inits=False,
             )
             wb.insert_workflow(wf_estmethod_original, predecessors=task_base_model_fit)
             candidate_no += 1
 
         # Create model with updated estimates from FOCE
         wf_estmethod_update = _create_candidate_model_wf(
-            candidate_no, method, solver, parameter_uncertainty_method, update_inits=True
+            candidate_no,
+            method,
+            solver,
+            parameter_uncertainty_method,
+            add_eval_after_est,
+            update_inits=True,
         )
         wb.insert_workflow(wf_estmethod_update, predecessors=task_base_model_fit)
         candidate_no += 1
@@ -63,6 +95,8 @@ def exhaustive_only_eval(methods, solvers, parameter_uncertainty_methods):
     task_start = Task('start', start)
     wb.add_task(task_start)
 
+    add_eval_after_est = len(methods) > 1
+
     candidate_no = 1
     for method, solver, parameter_uncertainty_method in itertools.product(
         methods, solvers, parameter_uncertainty_methods
@@ -72,6 +106,7 @@ def exhaustive_only_eval(methods, solvers, parameter_uncertainty_methods):
             method,
             solver,
             parameter_uncertainty_method,
+            add_eval_after_est,
             update_inits=False,
             only_evaluation=True,
         )
@@ -90,6 +125,7 @@ def _create_candidate_model_wf(
     method,
     solver,
     parameter_uncertainty_method,
+    add_eval_after_est,
     update_inits,
     only_evaluation=False,
 ):
@@ -111,6 +147,7 @@ def _create_candidate_model_wf(
         method,
         solver,
         parameter_uncertainty_method,
+        add_eval_after_est,
         update_inits,
         only_evaluation,
     )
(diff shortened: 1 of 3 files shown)


src/pharmpy/tools/estmethod/algorithms.py
```diff
@@ -146,22 +187,50 @@ def _create_base_model(model):
...
+        eval_settings = None
+        eval_method = None
+
+    if only_evaluation:
+        all_estrec_methods = [eval_method]
+    elif add_eval_after_est:
+        all_estrec_methods = [method, eval_method]
+    else:
+        all_estrec_methods = [method]
+
```
stellabelin, line 230: This logic is a bit confusing to me, especially in combination with the code on 243-250. So we have three possible cases: 

1. Only evaluation
2. Estimation and evaluation (relevant when assessing `methods`)
3. Only estimation (relevant when not assessing `methods`)

To me it would be a little easier to read if all of these things were set in one `if-else` block, since e.g. the description could be changed afterwards.

src/pharmpy/tools/estmethod/algorithms.py
```diff
@@ -171,31 +240,37 @@ def _create_candidate_model(
...
-def _create_est_settings(method, only_evaluation=False):
+def _create_est_settings(method, parameter_uncertainty_method=None, add_eval_after_est=False):
     est_settings = {
         'method': method,
         'interaction': True,
         'laplace': False,
         'auto': True,
         'keep_every_nth_iter': 10,
-        'parameter_uncertainty_method': None,
+        'maximum_evaluations': 9999,
```
stellabelin, line 263: How come this setting is added?

src/pharmpy/tools/estmethod/algorithms.py
```diff
@@ -171,31 +240,37 @@ def _create_candidate_model(
...
         est_settings['method'] = 'FOCE'
         est_settings['laplace'] = True
 
-    if only_evaluation:
-        est_settings['evaluation'] = True
+    if add_eval_after_est:
+        est_settings['parameter_uncertainty_method'] = None
     else:
-        est_settings['maximum_evaluations'] = 9999
+        est_settings['parameter_uncertainty_method'] = parameter_uncertainty_method
```
stellabelin, line 273: Why has this condition changed? Can't see where `evaluation=True`?

tests/tools/test_estmethod.py
```diff
@@ -33,30 +33,48 @@ def test_algorithm(algorithm, methods, solvers, parameter_uncertainty_methods, n
...
             'FO',
+            None,
+            True,
             '$ESTIMATION METHOD=ZERO INTER MAXEVAL=9999 AUTO=1 PRINT=10',
             '$ESTIMATION METHOD=IMP INTER EONLY=1 MAXEVAL=9999 ISAMPLE=10000 NITER=10 PRINT=10',
         ),
         (
             'LAPLACE',
+            None,
+            True,
```
stellabelin, line 48: Since both `parameter_uncertainty_method` and `add_eval_after_est` are the same for both tests I think it's clearer to explicitly set that as the option in the tool run and not have it parametrized.

stellabelin (changes requested):
Nice! I have some comments and questions.

src/pharmpy/tools/estmethod/algorithms.py
vrognas: Because the idea is that `est_settings` should always perform estimation, so MAXEVAL>0. If an evaluation is requested (which is the case for MAXEVAL=0), then we use the `eval_settings`. Currently only IMP evaluation is implemented, and not FOCE evaluation. Previously, evaluation was done both FOCE and then IMP in the same model, which makes no sense to me. I suggest we implement FOCE evaluation in a later PR.

vrognas: And `parameter_uncertainty_method` is allowed to be something else than `None` since we want to be able to just update that and not be forced to add an `IMP` evaluation step if we want to update the `parameter_uncertainty_method`

stellabelin: FOCE evaluation is supported here: 
https://github.com/pharmpy/pharmpy/blob/3008997a229bd0f492b1ce08775d544a5176bba8/src/pharmpy/model/external/nonmem/update.py#L1514-L1520
And tested for her:
https://github.com/pharmpy/pharmpy/blob/3008997a229bd0f492b1ce08775d544a5176bba8/src/pharmpy/model/external/nonmem/update.py#L1531-L1532
Basically the idea is that `evaluation=True` is independent of estimation method.

src/pharmpy/tools/estmethod/algorithms.py
vrognas: See comment https://github.com/pharmpy/pharmpy/pull/2068#discussion_r1333076138

src/pharmpy/tools/estmethod/algorithms.py
```diff
@@ -12,12 +12,19 @@ def exhaustive(methods, solvers, parameter_uncertainty_methods):
     task_start = Task('start', start)
     wb.add_task(task_start)
 
+    add_eval_after_est = True if methods is not None and len(methods) > 1 else False
```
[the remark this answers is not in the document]
rikardn: The ternary if is not needed. Simply use the boolean expression directly. `add_eval_after_test = len(methods) > 1` for example

src/pharmpy/tools/estmethod/algorithms.py
rikardn: `maximum_evaluations` is allowed to be `None` meaning whatever is the default (a "fair" amount according to the NONMEM userguide)

tests/tools/test_estmethod.py
vrognas: Fair enough! The original plan was to expand the test with more cases, which explains the parameters.

src/pharmpy/tools/estmethod/algorithms.py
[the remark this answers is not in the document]
vrognas: I thought it was easier to read this way (as a novice Python coder). But I can change it!

src/pharmpy/tools/estmethod/algorithms.py
```diff
@@ -122,21 +155,29 @@ def _copy_model(name, model):
...
+        eval_settings = None
+        eval_method = None
 
     base_model = model.replace(name='base_model')
-    est_method, eval_method = est_settings['method'], eval_settings['method']
 
-    if eval_method is not None:
-        parameter_uncertainty_method = eval_settings['parameter_uncertainty_method']
+    if add_eval_after_est:
+        all_estrec_methods = [est_method, eval_method]
```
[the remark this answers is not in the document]
vrognas: Changed!

src/pharmpy/tools/estmethod/algorithms.py
vrognas: Changed, see if the logic is easier to understand now.

src/pharmpy/tools/estmethod/algorithms.py
vrognas: Sure, FOCE evaluation is supported elsewhere in the Pharmpy code base, but in the `estmethod` tool it was never genuinely supported since an `IMP EONLY=1` step always followed.

src/pharmpy/tools/estmethod/algorithms.py
```diff
@@ -147,55 +188,87 @@ def _create_base_model(model):
...
+        est_settings = _create_est_settings(
+            method, parameter_uncertainty_method, add_eval_after_est
+        )
+        all_methods = [est_settings['method']]
+
+        if add_eval_after_est:
+            eval_settings = _create_eval_settings(laplace, parameter_uncertainty_method)
+            all_methods.append(eval_settings['method'])
+
+    # Update the model description
```
stellabelin, line 227: I think this comment can be removed.

src/pharmpy/tools/estmethod/algorithms.py
```diff
@@ -147,55 +188,87 @@ def _create_base_model(model):
...
         description=_create_description(
-            [method, eval_method],
+            methods=all_methods,
             solver=solver,
             parameter_uncertainty_method=parameter_uncertainty_method,
             update_inits=update_inits,
         )
     )
 
+    # Clear previous estimation steps
```
stellabelin, line 237: I think this comment can be removed.

src/pharmpy/tools/estmethod/algorithms.py
```diff
@@ -147,55 +188,87 @@ def _create_base_model(model):
...
     while len(model.estimation_steps) > 0:
         model = remove_estimation_step(model, 0)
 
-    model = add_estimation_step(model, **est_settings)
-    model = add_estimation_step(model, **eval_settings)
+    # Add new estimation steps
```
stellabelin, line 241: I think this comment can be removed.

src/pharmpy/tools/estmethod/algorithms.py
```diff
@@ -147,55 +188,87 @@ def _create_base_model(model):
...
+    if est_settings:
+        model = add_estimation_step(model, **est_settings)
+    if eval_settings:
+        model = add_estimation_step(model, **eval_settings)
+
+    # Update the solver, if provided
```
stellabelin, line 247: I think this comment can be removed.

stellabelin (changes requested):
Some more comments (most are related to comments).

src/pharmpy/tools/estmethod/algorithms.py
```diff
@@ -28,7 +35,14 @@ def exhaustive(methods, solvers, parameter_uncertainty_methods):
 def exhaustive_with_update(methods, solvers, parameter_uncertainty_methods):
     wb = WorkflowBuilder()
 
-    task_base_model = Task('create_base_model', _create_base_model)
+    add_eval_after_est = len(methods) > 1
+
+    task_base_model = Task(
+        'create_base_model',
+        _create_base_model,
+        parameter_uncertainty_methods[0],
```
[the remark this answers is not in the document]
vrognas: Good point about questioning the use of `parameter_uncertainty_methods[0]` for the base model. I chose this because that index is always guaranteed to be available, and it's efficient to use one of the `parameter_uncertainty_methods` that we're already testing. This way, we avoid running extra models that aren't part of the experiments. For example, if we're testing `IMP` and `FOCE` methods along with `SANDWICH` and `CPG` for parameter uncertainty, choosing `None` would add an extra run with `FOCE` and no parameter uncertainty—this isn't efficient.

However, there's a case I hadn't considered. If `FOCE` isn't in the search space and `parameter_uncertainty_methods` isn't `None`, then it's unnecessary to use one of the parameter uncertainty methods for the base model. The base model is for estimation, not for evaluating parameter uncertainty. So, I've added some logic to handle this scenario:

```python
parameter_uncertainty_method_base = parameter_uncertainty_methods[0] if 'FOCE' in methods else None
```

This way, if FOCE isn't part of the search space, the base model won't use any of the parameter uncertainty methods, which makes more sense. Thanks for bringing this to my attention!

src/pharmpy/tools/estmethod/algorithms.py
```diff
@@ -122,21 +155,29 @@ def _copy_model(name, model):
...
     base_model = model.replace(name='base_model')
-    est_method, eval_method = est_settings['method'], eval_settings['method']
 
-    if eval_method is not None:
-        parameter_uncertainty_method = eval_settings['parameter_uncertainty_method']
+    if add_eval_after_est:
+        all_methods = [est_method, eval_method]
     else:
-        parameter_uncertainty_method = est_settings['parameter_uncertainty_method']
+        all_methods = [est_method]
```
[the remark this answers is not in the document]
vrognas: Indeed. Revised logic.

stellabelin (approved):
LGTM!

Issue #2004, opened 2023-08-31
Support `$DESIGN` as an uncertainty method


vrognas:
### Feature request

Support `$DESIGN` as an uncertainty method.

vrognas:
Initial support was added via [#2004](https://github.com/pharmpy/pharmpy/issues/2004)

PR #2077, opened 2023-09-22
Add `$DESIGN` (EFIM) support

vrognas:
This PR adds EFIM (Expected Fisher Information Matrix) support.
Add 'EFIM' to `supported_parameter_uncertainty_methods` in `estimation.py`.
Introduce a new function `add_efim_records` that manipulates the control stream for EFIM calculations.

Resolves Issue #2004

### DIFF
--- a/src/pharmpy/model/external/nonmem/nmtran_parser.py
+++ b/src/pharmpy/model/external/nonmem/nmtran_parser.py
@@ -89,14 +89,14 @@ class NMTranControlStream:
         self._active_problem = 0
         self.abbreviated = Abbreviated(self)
 
-    def get_records(self, name):
+    def get_records(self, name, problem_no=0):
         """Return a list of all records of a certain type in the current $PROBLEM"""
         current_problem = -1
         found = []
         for record in self.records:
             if record.name == 'PROBLEM':
                 current_problem += 1
-            if current_problem == self._active_problem and record.name == name:
+            if current_problem == problem_no and record.name == name:
                 found.append(record)
         return found
 
--- a/src/pharmpy/model/external/nonmem/update.py
+++ b/src/pharmpy/model/external/nonmem/update.py
@@ -1395,7 +1395,7 @@ def add_parameters_ratio(model: Model, numpar, denompar, source, dest):
 def define_parameter(
     model: Model, name: str, value: sympy.Expr, synonyms: Optional[List[str]] = None
 ):
-    """Define a parameter in statments if not defined
+    """Define a parameter in statements if not defined
     Update if already defined as other value
     return True if new assignment was added
     """
@@ -1480,6 +1480,44 @@ def update_abbr_record(model: Model, rv_trans):
 
     model = model.replace(internals=model.internals.replace(control_stream=control_stream))
     return model, trans
+
+
+def add_efim_records(control_stream, idx_cov, last_est_rec):
+    data_record = str(control_stream.get_records("DATA")[0]).rstrip()
+    input_record = str(control_stream.get_records("INPUT")[0]).rstrip()
+    efim_records = [
+        "$PROBLEM DESIGN",
+        f"{data_record} REWIND",
+        f"{input_record}",
+        "$MSFI efim.msf",
+        "$DESIGN APPROX=FO FIMDIAG=1 GROUPSIZE=1 OFVTYPE=1",
+    ]
+
+    for i, rec in enumerate(efim_records, start=1):
+        new_record = create_record(f"{rec}\n")
+        control_stream = control_stream.insert_record(new_record, at_index=idx_cov + i)
+
+    last_est_rec_msfo = last_est_rec.append_option("MSFO", "efim.msf")
+    control_stream = control_stream.replace_records([last_est_rec], [last_est_rec_msfo])
+
+    return control_stream
+
+
+def add_covariance_record(control_stream, idx_cov, method):
+    method_to_cov = {
+        "SANDWICH": "$COVARIANCE UNCONDITIONAL PRINT=E PRECOND=1",
+        "CPG": "$COVARIANCE MATRIX=S UNCONDITIONAL PRINT=E PRECOND=1",
+        "OFIM": "$COVARIANCE MATRIX=R UNCONDITIONAL PRINT=E PRECOND=1",
+    }
+
+    try:
+        covrec_str = method_to_cov[method]
+        covrec = create_record(covrec_str + "\n")
+        control_stream = control_stream.insert_record(covrec, at_index=idx_cov + 1)
+    except KeyError:
+        raise ValueError(f"Unsupported parameter uncertainty method: {method}")
+
+    return control_stream
 
 
 def update_estimation(control_stream, model):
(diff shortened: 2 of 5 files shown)


src/pharmpy/model/external/nonmem/update.py
```diff
@@ -1482,6 +1482,45 @@ def update_abbr_record(model: Model, rv_trans):
...
+        f"{data_record} REWIND",
+        f"{input_record}",
+        "$MSFI efim.msf",
+        "$DESIGN APPROX=FO FIMDIAG=1 GROUPSIZE=1 OFVTYPE=1",
+    ]
+
+    # EFIM records need to be added in reverse order to the control stream
+    new_records = [create_record(f"{rec}\n") for rec in efim_records]
+    for new_record in reversed(new_records):
+        control_stream = control_stream.insert_record(new_record, at_index=idx_cov + 1)
```
stellabelin, line 1499: This doesn't need to be done in two steps right? If you just loop through `efim_records` you can create and insert directly, i.e.:

```
for efim_rec in efim_recrods:
   new_record = create_record(f"{rec}\n")
   control_stream = control_stream.insert_record(new_record, at_index=idx_cov + 1)
```

src/pharmpy/model/external/nonmem/update.py
```diff
@@ -1482,6 +1482,45 @@ def update_abbr_record(model: Model, rv_trans):
...
+
+    # MSFO in $ESTIMATION needed for EFIM in next $PROBLEM
```
stellabelin, line 1501: I think this comment is perhaps not necessary, from the code we can see that MSFO is added and we know it's needed since we're adding it.

src/pharmpy/model/external/nonmem/update.py
```diff
@@ -1584,21 +1623,30 @@ def update_estimation(control_stream, model):
...
+        else:
+            control_stream = add_covariance_record(
+                control_stream, idx_cov, new_parameter_uncertainty_method
+            )
+
     elif old_parameter_uncertainty_method is not None and new_parameter_uncertainty_method is None:
-        # Remove $COVARIANCE
-        covrecs = control_stream.get_records('COVARIANCE')
-        control_stream = control_stream.remove_records(covrecs)
+        # If old_parameter_uncertainty_method is EFIM, remove EFIM records
```
stellabelin, line 1638: I think this comment can be removed

src/pharmpy/model/external/nonmem/update.py
```diff
@@ -1584,21 +1623,30 @@ def update_estimation(control_stream, model):
...
+        if old_parameter_uncertainty_method == "EFIM":
+            record_types = ["DESIGN", "MSFI", "INPUT", "DATA", "PROBLEM"]
+
+            for rec_type in record_types:
+                last_rec = control_stream.get_records(rec_type, all_problems=True)[-1]
+                control_stream = control_stream.remove_records([last_rec])
+
+        # If old_parameter_uncertainty_method is not EFIM, remove COVARIANCE records
```
stellabelin, line 1646: I think this comment can be removed.

src/pharmpy/model/external/nonmem/update.py
```diff
@@ -1584,21 +1623,30 @@ def update_estimation(control_stream, model):
...
+                control_stream, idx_cov, new_parameter_uncertainty_method
+            )
+
     elif old_parameter_uncertainty_method is not None and new_parameter_uncertainty_method is None:
-        # Remove $COVARIANCE
-        covrecs = control_stream.get_records('COVARIANCE')
-        control_stream = control_stream.remove_records(covrecs)
+        # If old_parameter_uncertainty_method is EFIM, remove EFIM records
+        if old_parameter_uncertainty_method == "EFIM":
+            record_types = ["DESIGN", "MSFI", "INPUT", "DATA", "PROBLEM"]
```
stellabelin, line 1640: Why is this in reverse order?

stellabelin (changes requested):
Some comments. The main one is regarding the new option to `get_records`.

src/pharmpy/model/external/nonmem/nmtran_parser.py
```diff
@@ -89,15 +89,23 @@ def __init__(self, records=None):
         self._active_problem = 0
         self.abbreviated = Abbreviated(self)
 
-    def get_records(self, name):
-        """Return a list of all records of a certain type in the current $PROBLEM"""
-        current_problem = -1
+    def get_records(self, name, all_problems=False):
```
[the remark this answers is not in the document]
vrognas: Perhaps problem_no could be either an integer (specifying the problem number) or 'all'?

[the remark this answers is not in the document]
stellabelin: I think it's more intuitive if it's always an index (that the same variable can't have multiple types), what would be the use case for getting all at the same time?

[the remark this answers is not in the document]
vrognas: I see where you're coming from. However, the current design uses `all_problems=True` to fetch records from all problems in the model file. For instance, in this [GitHub discussion](https://github.com/pharmpy/pharmpy/pull/2077#discussion_r1335565851), `get_records` is used to get all records, not just from a part of the model. Doesn't it make sense for a function called `get_records` to return all records by default?

The use case for fetching all records would be when you want to look at or manipulate the entire model file, not just a part of it. For example, in some operations, you might need to update records across multiple problems in the model. A simple example would be if you change the name of your input data file.

If we strictly use an integer index for `problem_no`, we'd need to store the number of problems somewhere, like in an array(?), to reference it. I'm not entirely sure where the best place to implement that would be. Do you have any suggestions?

src/pharmpy/model/external/nonmem/update.py
vrognas: Because $PROBLEM needs to be removed last, since we are using that to be sure we only remove records from the last $PROBLEM

src/pharmpy/model/external/nonmem/nmtran_parser.py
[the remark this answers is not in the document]
vrognas: Force-pushed changes as per our offline discussion

src/pharmpy/model/external/nonmem/update.py
```diff
@@ -1482,6 +1482,44 @@ def update_abbr_record(model: Model, rv_trans):
...
+    input_record = str(control_stream.get_records("INPUT")[0]).rstrip()
+    efim_records = [
+        "$PROBLEM DESIGN",
+        f"{data_record} REWIND",
+        f"{input_record}",
+        "$MSFI efim.msf",
+        "$DESIGN APPROX=FO FIMDIAG=1 GROUPSIZE=1 OFVTYPE=1",
+    ]
+
+    for i, rec in enumerate(efim_records):
```
stellabelin, line 1496: If you do `enumerate(efim_records, 1)` it will start at 1 and will not have to be added in `insert_record`!

stellabelin (approved):
LGTM! Small comment

src/pharmpy/model/external/nonmem/update.py
vrognas: Very true! Great catch 🎣. Force pushed.

PR #2078, opened 2023-09-22
Change naming of bolus absorption and add verb get_ to number_of

JohanWBorg:
In response to the final comments of #2032.
- Change name of bolus absorption to instantaneous absorption
- Add verb _get to number_of_peripheral/transit_compartments
- Added has_ absorption type and number_of functions to /modeling __innit__ file


src/pharmpy/modeling/odes.py
```diff
@@ -1505,7 +1505,7 @@ def has_first_order_absorption(model: Model):
 
     """
 
-    odes = model.statements.ode_system
+    odes = get_and_check_odes(model)
     if odes is None:
```
rikardn, line 1509: Now this is not needed. `get_and_check_odes` guarantees the return of a `CompartmentalSystem` (else it raises)

rikardn (changes requested):
One comment

src/pharmpy/modeling/odes.py
JohanWBorg: Oh sorry, I'll fix it right away!

rikardn:
Super!

PR #2087, opened 2023-09-26
Add test model for `$DESIGN`

vrognas:
Prepare testdata for integration tests of control streams with $DESIGN as parameter uncertainty method.


rikardn:
Nice!

PR #2098, opened 2023-09-27
Correct default search space in AMD

JohanWBorg:
After the MFL update, search spaces need to be explicit and so the defualt searchspaces within AMD have been updated to match this new way of thinking


rikardn:
Thanks!

Issue #2060, opened 2023-09-19
error from add_periphral_compartment 


keqiss:
### Bug report

When the dataset and model have CMT column even `CMT = DROP`,  `add_periphral_compartment`  sometimes fails which interrupts AMD run and gives errors as shown. When CMT in dataset and model file is deleted then there wouldn't be an error. 

Pharmpy version: 0.101.2

```
2023-09-18 13:18:01,700 - distributed.worker - WARNING - Compute Failed
Key:       PERIPHERALS(1)-add_iivs-choose_best_model-copy-update_inits-run0-2fd0e6fc-e8ea-4e46-b8f9-11a386a3d6b7
Function:  execute_task
args:      ((<function retrieve_from_database_or_execute_model_with_tool.<locals>.task at 0x2b5e08988180>, <pharmpy.workflows.tool_database.local_directory.LocalDirectoryToolDatabase object at 0x2b5dead80910>, (<function _add_iiv_to_func at 0x2b5dea55f380>, 'absorption_delay', (<function _apply_transformation at 0x2b5dea55f100>, ('PERIPHERALS', 1), functools.partial(<function set_peripheral_compartments at 0x2b5dc57b3420>, n=1), (<function update_initial_estimates at 0x2b5dea390180>, (<function _copy at 0x2b5dea55f2e0>, 'modelsearch_run109', (('PERIPHERALS', 1),), (<function _get_best_model at 0x2b5dea55ee80>, <Pharmpy model object modelsearch_run37>, <Pharmpy model object modelsearch_run56>)))))))
kwargs:    {}
Exception: "KeyError('TRANSIT1')"

<Client: 'inproc://172.18.5.6/9736/1' processes=1 threads=16, memory=125.00 GiB>
<Client: 'inproc://172.18.5.6/9736/10' processes=1 threads=16, memory=125.00 GiB>
Traceback (most recent call last):
  File "/crex/proj/pmgroup_active_storage/private/users/keqi/AMDTEST/testrun_mbic2/moxonidine/amdtest.py", line 10, in <module>
    res = run_amd(input = model,
          ^^^^^^^^^^^^^^^^^^^^^^
  File "/domus/h1/kesh5913/pharmpy/src/pharmpy/tools/amd/run.py", line 266, in run_amd
    subresults = func(next_model)
                 ^^^^^^^^^^^^^^^^
[...]

PR #2105, opened 2023-09-27
Update to automatic CMT column update

JohanWBorg:
Update CMT column in response to issue #2060. Will now only update if needed. In other cases (as with a previously dropped CMT column) a new CMT column will be generated, overwriting the previous one.

### DIFF
--- a/src/pharmpy/modeling/data.py
+++ b/src/pharmpy/modeling/data.py
@@ -990,13 +990,6 @@ def get_cmt(model: Model):
     else:
         return model.dataset[cmtcols[0].name]
 
-    try:
-        cmtcols = model.dataset["CMT"]
-    except KeyError:
-        pass
-    else:
-        return cmtcols
-
     # See if admid exist
     try:
         admidcols = di.typeix["admid"]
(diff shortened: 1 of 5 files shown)


src/pharmpy/modeling/data.py
```diff
@@ -964,38 +964,43 @@ def set_admid(model: Model, column_name: str):
...
         CMT
     """
     di = model.datainfo
-    try:
-        cmtcols = di.typeix['compartment']
-    except IndexError:
-        pass
-    else:
-        return model.dataset[cmtcols[0].name]
+    if not force:
```
rikardn, line 990: What should `force` do? I don't see any code for that case. Is it necessary? Perhaps `get_cmt` should create the column if it doesn't exist (not adding it to the dataset though).

JohanWBorg: The idea behind force is to forcefully create a new CMT column (even if a previous one exists), as written in the docstring. All code needed for this is the `if not force:` statement. This functionality is used by the CMT update machinery to update previously dropped CMT columns which has since become outdated.
As `get_cmt()` extracts based on firstly the `compartment` datainfo type and secondly the name 'CMT', I thought this could be useful in this function as well. Since pharmpy will still save "CMT" columns in the dataset even if dropped, that column would be returned by `get_cmt()` (even if the user might not be interested in that one)

PR #2111, opened 2023-09-28
Refactor `_parse_table`

vrognas:
Improved variable naming for better readability (e.g., `m` to `table_number_match`)

### DIFF
--- a/src/pharmpy/model/external/nonmem/table.py
+++ b/src/pharmpy/model/external/nonmem/table.py
@@ -80,21 +80,21 @@ class NONMEMTableFile:
             table.is_evaluation = False
             if re.search(r'(Evaluation)', table_title_line):
                 table.is_evaluation = True  # No estimation step was run
-            m = re.match(
+            table_attrs_match = re.match(
                 r'TABLE NO.\s+\d+: (.*?): (?:Goal Function=(.*): )?Problem=(\d+) '
                 r'Subproblem=(\d+) Superproblem1=(\d+) Iteration1=(\d+) Superproblem2=(\d+) '
                 r'Iteration2=(\d+)',
                 table_title_line,
             )
-            if m:
-                table.method = m.group(1)
-                table.goal_function = m.group(2)
-                table.problem = int(m.group(3))
-                table.subproblem = int(m.group(4))
-                table.superproblem1 = int(m.group(5))
-                table.iteration1 = int(m.group(6))
-                table.superproblem2 = int(m.group(7))
-                table.iteration2 = int(m.group(8))
+            if table_attrs_match:
+                table.method = table_attrs_match.group(1)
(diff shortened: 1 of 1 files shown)


src/pharmpy/model/external/nonmem/table.py
```diff
@@ -59,7 +59,7 @@ def _parse_table(
     ) -> NONMEMTable:
         # NOTE: Content lines must contain endlines!
 
-        table_line = None if notitle else content.pop(0)
+        table_title_line = None if notitle else content.pop(0)
```
rikardn, line 62: It is the line containing "TABLE" so I don't think this change makes it clearer

src/pharmpy/model/external/nonmem/table.py
```diff
@@ -73,10 +73,10 @@ def _parse_table(
             table = NONMEMTable(''.join(content))  # Fallback to non-specific table type
 
         if table_title_line is not None:
-            m = re.match(r'TABLE NO.\s+(\d+)', table_title_line)
-            if not m:
+            table_number_match = re.match(r'TABLE NO.\s+(\d+)', table_title_line)
```
rikardn, line 76: The name `m` for a match is standard in Python so this change makes things less clear in my opinion.

src/pharmpy/model/external/nonmem/table.py
```diff
@@ -80,21 +80,21 @@ def _parse_table(
             table.is_evaluation = False
             if re.search(r'(Evaluation)', table_title_line):
                 table.is_evaluation = True  # No estimation step was run
-            m = re.match(
+            table_attrs_match = re.match(
```
rikardn, line 83: Same thing here. `m` is clearer because it is a de facto standard

src/pharmpy/model/external/nonmem/table.py
```diff
@@ -65,7 +65,7 @@ def _parse_table(
             table = ExtTable(''.join(content))
         elif suffix == '.phi':
             table = PhiTable(''.join(content))
-        elif suffix == '.cov' or suffix == '.cor' or suffix == '.coi':
+        elif suffix in ['.cov', '.cor', '.coi']:
```
rikardn, line 68: To me either way is fine so better to keep the original code

src/pharmpy/model/external/nonmem/table.py
```diff
@@ -209,7 +209,7 @@ class PhiTable(NONMEMTable):
     @property
     def iofv(self):
         df = self._df.copy(deep=True)
-        # Remove all zero lines. These are from individuals with no observations
+        # Remove all zero lines. These are from individuals with no observations.
```
rikardn, line 212: Touching a line of code change `git blame` obscuring the original author of the comment. So for little added value I would like to keep this without changes

PR #2115, opened 2023-09-28
Updated MFL- and modelsearch docs

JohanWBorg:
Update the MFL documentation and added instantaneous absorption type as well as ON/OFF arguments for lagtime.
For the modelsearch documentation, I added information regarding the new workflow logic regarding the addition of a base model and explicit search spaces.

I would love some feedback regarding the modelsearch workflow to make sure it is clear and understandable :)

### DIFF
--- a/docs/amd.rst
+++ b/docs/amd.rst
@@ -113,11 +113,11 @@ default is:
 
 .. code-block::
 
-    ABSORPTION([ZO,SEQ-ZO-FO])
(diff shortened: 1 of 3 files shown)


rikardn:
Thanks! I'll merge without reviewing in detail. At some point we could review the user guides by looking at the html. There is also an issue to reduce the default scope where we need to change some documentation.

PR #2123, opened 2023-09-29
Enable `_parse_modelfit_results` to parse models with `$DESIGN`

vrognas:
This PR allows us to parse modelfit results from models that use the EFIM ($DESIGN) parameter uncertainty method. Additional relevant tests cases for the correct parsing of RSE has been added. Also updated names from "covariance" to "parameter_uncertainty" to emphasize that we now also support `$DESIGN` and not only `$COVARIANCE` for parameter uncertainty estimation.


rikardn:
Closing in favour of #2199

PR #2148, opened 2023-10-03
Only index models with results not None in summary_models

JohanWBorg:
Fix an issue will summarizing the results in modelsearch where models with results "None" tried to be indexed in the table "summary_models"

### DIFF
--- a/src/pharmpy/tools/modelsearch/tool.py
+++ b/src/pharmpy/tools/modelsearch/tool.py
@@ -258,8 +258,8 @@ def post_process(rank_type, cutoff, *models):
         results_to_summarize.extend(model.modelfit_results for model in res_models)
 
     summary_models = summarize_modelfit_results(results_to_summarize)
-    summary_models['step'] = [0] + [1] * (len(results_to_summarize) - 1)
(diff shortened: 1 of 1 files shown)


src/pharmpy/tools/modelsearch/tool.py
```diff
@@ -258,8 +258,12 @@ def post_process(rank_type, cutoff, *models):
         results_to_summarize.extend(model.modelfit_results for model in res_models)
 
     summary_models = summarize_modelfit_results(results_to_summarize)
-    summary_models['step'] = [0] + [1] * (len(results_to_summarize) - 1)
-    summary_models.reset_index().set_index(['step', 'model'])
+    summary_models.insert(
```
rikardn, line 261: Could this be replaced with `summary_models['step'] = [0] + [1] * (len(summary_models) - 1)` ? If that works it might be a bit simpler.

JohanWBorg: Oooh, did not think of that. I'll check it out and get back to you

JohanWBorg: Implemented your suggestion :)

PR #2182, opened 2023-10-10
Drug metabolite default search space and correct candidate models

JohanWBorg:
Drug metabolite models will now run with the default modelsearch search space for the parent drug metabolite
Candidate models is also dependent on the administration route and presystemic drug metabolite models are only created for oral/ivoral administration


stellabelin (approved):
LGTM!

PR #2196, opened 2023-10-12
Covsearch update/fix

JohanWBorg:
Updates/fixes for covsearch
- Correct use of parent model in order to correctly compute reduced OFV values
- Correct numbering of steps if last forward step failed before going to backwards search
- Return correct final model from model search instead of final model from forward search
- Correct reduced_ofv and extended_ofv 
- Added p-value for backwards search models instead of NaN


src/pharmpy/tools/covsearch/tool.py
```diff
@@ -479,38 +497,63 @@ def _make_df_steps(best_model: Model, candidates: List[Candidate]):
...
+            reduced_ofv = np.nan if (mfr := parent_model.modelfit_results) is None else mfr.ofv
+            extended_ofv = np.nan if (mfr := model.modelfit_results) is None else mfr.ofv
+            alpha = last_step.alpha
+            extended_significant = lrt_test(
+                parent_model,
+                candidate.model,
+                reduced_ofv,
+                extended_ofv,
+                alpha,
+            )
```
stellabelin, line 524: `ofv_drop` needs to be defined here too right?

src/pharmpy/tools/covsearch/tool.py
```diff
@@ -467,8 +470,23 @@ def _make_df_steps(best_model: Model, candidates: List[Candidate]):
     models_dict = {candidate.model.name: candidate.model for candidate in candidates}
     children_count = Counter(candidate.model.parent_model for candidate in candidates)
 
+    # Find if longest forward is also the input for the backwards search
+    # otherwise add an index offset to _make_df_steps_function
+    only_forward = [
+        fc for fc in candidates if fc.steps == tuple() or isinstance(fc.steps[-1], ForwardStep)
+    ]
+    largest_forward_step = max([len(fc.steps) for fc in only_forward])
+    only_forwards_largest = [fc for fc in only_forward if len(fc.steps) == largest_forward_step]
```
stellabelin, line 479: I think this part is a bit confusing, I don't know if it would help a bit if some variables are renamed? E.g. `only_forwards` -> `forwards_candidates`, `only_forwards_largest` -> `largest_candidates`? Not sure if it helps, to me it's just a bit unclear what kind of objects these lists consist of.

stellabelin (commented):
I think this looks correct, one variable that might need to be defined (see comment) and thought one part was slightly confusing (but maybe that's unavoidable in this case).

src/pharmpy/tools/covsearch/tool.py
JohanWBorg: `ofv_drop` is defined after if statements on line 536

src/pharmpy/tools/covsearch/tool.py
JohanWBorg: renamed the two following your suggestions with a slight twist.
`only_forwards` -> `forward_candidates`
`only_forwards_largest` -> `largest_forward_candidates`

stellabelin (approved):
LGTM!

PR #2215, opened 2023-10-18
COVsearch update

JohanWBorg:
Implemented covariate as a part of ModelFeatures and changes to covsearch for handling of already present covariates
- Covariate attribute is now a part of ModelFeautures, allowing addition, subtraction and least number of transformation between different models/search spaces
- Possibility of extracting covariates from a model of choosing, with effect (based on our effect templates) and operator using "get_covariates()". Also implemented this feature within "get_model_features()"
- Changes to covsearch to handle already present covariates. If present covariates is not part of search space, they are removed before running covsearch. If they are a part of the model and the search space, they are left as is and are removed from the search space.

### DIFF
--- a/docs/covsearch.rst
+++ b/docs/covsearch.rst
@@ -120,6 +120,9 @@ For instance,
 would test an exponential covariate effect on clearance and volume for each
 continuous covariate.
 
+.. note::
+    Covariates that are already present in the model will be removed, unless they are also part of the search space. See :ref:`Algorithm<algorithm_covsearch>` for more.
+
 More automatic symbols are available. They are described in the :ref:`MFL
 symbols section<mfl_symbols>`.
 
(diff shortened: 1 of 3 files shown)


rikardn:
@JohanWBorg Could you have a look at the conflicts?

JohanWBorg:
@rikardn Conflicts should be resolved now

src/pharmpy/modeling/covariate_effect.py
```diff
@@ -14,13 +15,116 @@
...
+    remove_covariate_effect_from_statements,
+    simplify_model,
+)
+from .parameters import get_thetas
 
 EffectType = Union[Literal['lin', 'cat', 'piece_lin', 'exp', 'pow'], str]
 OperationType = Literal['*', '+']
 
 
+def get_covariates(model: Model):
```
rikardn, line 32: Need return type hint for the pharmr code generation

src/pharmpy/tools/mfl/parse.py
```diff
@@ -703,13 +707,31 @@ def get_model_features(model: Model, supress_warnings: bool = False) -> str:
         transits = f'TRANSITS({transits}{","+depot})'
     if peripherals != 0:
         peripherals = f'PERIPHERALS({peripherals})'
+    if len(covariates) != 0:
+        # if any([v[0][0] == 'custom' for _, v in covariates.items()]):
```
rikardn, line 711: Remove commented code

rikardn (changes requested):

PR #2219, opened 2023-10-19
Update metabolite pk param extraction to use function for amounts

JohanWBorg:
Should fix failing unit test for test_get_pk_parameters_metabolite


rikardn:
LGTM

PR #2244, opened 2023-10-27
Use warnings instead of print for rxode/nlmixr warnings

JohanWBorg:


stellabelin (commented):
To make the code more concise, you can use: `with pytest.warns()` (see https://github.com/pharmpy/pharmpy/blob/74e5ec80335fc5129a1f26f48a40ed7913d25962/tests/nonmem/test_nonmem_model.py#L413)

stellabelin (approved):
LGTM!

PR #2284, opened 2023-11-07
Fix crashing modelsearch due to wrongfully connected modelfit_results object

JohanWBorg:
Previoulsy connected modelfit_results objects will now be removed when using ModelEntry (if connected)

### DIFF
--- a/src/pharmpy/tools/external/nonmem/run.py
+++ b/src/pharmpy/tools/external/nonmem/run.py
@@ -22,6 +22,11 @@ PARENT_DIR = f'..{os.path.sep}'
 def execute_model(model_or_model_entry, db):
     if isinstance(model_or_model_entry, ModelEntry):
         model = model_or_model_entry.model
+
+        # TODO : Remove after complete integration of ModelEntry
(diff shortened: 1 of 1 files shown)


src/pharmpy/tools/external/nonmem/run.py
```diff
@@ -22,6 +22,11 @@
 def execute_model(model_or_model_entry, db):
     if isinstance(model_or_model_entry, ModelEntry):
         model = model_or_model_entry.model
+
+        # TODO : Remove after complete integration of ModelEntry
+        if model.modelfit_results is None:
+            model = model.replace(modelfit_results=None)
```
rikardn, line 28: If `modelfit_results` was checked to be `None` why set it to `None` again?

JohanWBorg: Oh no this is incorrect. The if statement was added to avoid performing the operations when they aren't needed but it should be `if model.modelfit_results is not None:`
Great catch, I'll fix it right away

PR #2293, opened 2023-11-08
Avoid removing allometry scaling in covsearch

JohanWBorg:
Added allometric scaling effects are now extracted and added to the covsearch search space in order to be left in the model when allometry is performed before covsearch

Also implemented better determination technique for covariate effects in order to more accurately be able to determine the effect type in more complex expressions.

### DIFF
--- a/src/pharmpy/modeling/allometry.py
+++ b/src/pharmpy/modeling/allometry.py
@@ -87,6 +87,8 @@ def add_allometry(
     S₁ = V
 
     """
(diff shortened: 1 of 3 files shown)


rikardn:
This looks good. Some conflicts snuck in though.

JohanWBorg:
@rikardn conflicts should be resolved now :)

rikardn:
Thanks!

PR #2294, opened 2023-11-08
Default allometric variable based on "body weight" instead of using "WT"

JohanWBorg:
When adding allometric scaling. If no allometric variable is defined, try to extract one from the dataset based on "body weight" descriptor instead of using "WT" as the default value

### DIFF
--- a/src/pharmpy/modeling/allometry.py
+++ b/src/pharmpy/modeling/allometry.py
@@ -16,7 +16,7 @@ else:
 
 def add_allometry(
     model: Model,
-    allometric_variable: Union[str, sympy.Expr] = 'WT',
+    allometric_variable: Optional[Union[str, sympy.Expr]] = None,
     reference_value: Union[str, int, float, sympy.Expr] = 70,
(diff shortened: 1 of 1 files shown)


src/pharmpy/modeling/allometry.py
```diff
@@ -84,6 +87,15 @@ def add_allometry(
     S₁ = V
 
     """
+    if allometric_variable is None:
+        for col in model.datainfo:
+            if col.descriptor == 'body weight':
+                allometric_variable = col.name
```
rikardn, line 93: There is a nice `else' for loops that could be used here.

```python
for col in model.datainfo:
    if col.descriptor == 'body weight':
        allometric_variable = col.name
        break
else:
    raise ValueError(...)
```

Even better would be to use the `descriptorix` of `DataInfo` instead of looping

JohanWBorg: Thanks for the comment! I redid the change using `descriptorix` as per your suggestion. Let me know if there are any other comments :)

rikardn:
Nice!

PR #2303, opened 2023-11-09
File path for windows in nlmixr/rxode

JohanWBorg:
Fixed issue with file path when running models in R.
Also fixes conversion issue by changing A_CENTRAL(t) variables to A_CENTRAL due to function naming (A_CENTRAL(t)) not being supported in nlmixr/rxode


rikardn:
I think "tox -e format" should solve the issue

PR #2307, opened 2023-11-10
Correct windows path for fix_eta file and updates to RxODE

JohanWBorg:
Fix backslash issue in file path for windows when running RxODE and nlmixr. Also added support for running fit(model, "rxode") and similar updates to RxODE as for nlmixr in #2303


rikardn:
LGTM

PR #2369, opened 2023-11-22
Rework covsearch for structural/mechanistic and exploratory covariates

JohanWBorg:
Updates to covsearch for handling of optional and mandatory covariates using the COVARIATE(...) vs COVARIATE?(...) notation
Also added possibility of defining mechanistic covariates as argument for AMD which would cause AMD to run a initial covsearch run with the specified covariates.
ModelEntry has also now been implemented for covsearch as a part of this update


rikardn:
LGTM

PR #2393, opened 2023-11-29
Conflict resolving (after covsearch, allometry and strictness update)

JohanWBorg:


src/pharmpy/modeling/covariate_effect.py
```diff
@@ -47,6 +47,7 @@ def get_covariates(model: Model) -> dict[list]:
     """
     parameters = get_individual_parameters(model)
     covariates = get_model_covariates(model)
+    covariates = [c for c in covariates if model.datainfo[str(c)].type == "covariate"]
```
rikardn, line 50: Why is it necessary for a covariate to be marked as such in the datainfo?

JohanWBorg: If it is not marked as a covariate, then other data columns could be interpreted as being covariates and could therefore be removed in COVsearch (as they are not part of the search space)
For instance, if in IOVsearch, an occasion is added (for example "VISI" with type "unknown"), this will be interpreted as a covariate and its effect would be removed in COVsearch (if we don't filter on "covariate" type as implemented).

rikardn: I just realized that this is not in `set_covariate_effect`. My mistake.

PR #2402, opened 2023-12-01
Write comparison file when comparing verifying RxODE

JohanWBorg:


rikardn:
Do we need to write it? Could it be returned from the function instead?

JohanWBorg:
It is possible to return it as well but for ease of storing past results and to give the possibility of looking at a file instead (and to follow the same procedure as for nlmixr verification) I thought this would be a nice feature. Naturally, this could be done by the user instead.

PR #2412, opened 2023-12-04
Retries tool

JohanWBorg:
Creation of a new Retries tool, for rerunning a model with tweaked initial parameter estimates. 
The tool is also implemented in AMD, allowing the user to run the retries tool on only the final AMD model, the resulting model from each tool, or skip the retries tool altogether.


src/pharmpy/tools/retries/tool.py
```diff
@@ -0,0 +1,274 @@
...
+
+    Parameters
+    ----------
+    model : Optional[Model], optional
+        Model object to run retries on. The default is None.
+    results : Optional[ModelfitResults], optional
+        Connected ModelfitResults object. The default is None.
+    number_of_candidates : int, optional
+        Number of retry candidates to run. The default is 5.
+    allowed_variance: float
```
rikardn, line 45: The name of `allowed_variance`  should be changed. It is not a variance and allowed is kind of implicit. I have tried to find a good word for this, but I only have some suggestions. `degree` is what PsN is using. Otherwise `range` or `interval`.

src/pharmpy/tools/amd/run.py
```diff
@@ -66,6 +66,7 @@ def run_amd(
     strictness: Optional[str] = "minimization_successful or (rounding_errors and sigdigs>=0.1)",
     dv_types: Optional[dict] = None,
     mechanistic_covariates: Optional[List[str]] = None,
+    perform_retries: Optional[str] = "amd_final",
```
rikardn, line 69: `perform_retries` sounds like a boolean. Perhaps only `retries` or `retries_strategy`.

src/pharmpy/tools/amd/run.py
```diff
@@ -138,6 +142,9 @@ def run_amd(
     run_tool
 
     """
+    # NORMAL SCALE
+    # UCP SCALE
```
rikardn, line 146: Remove

src/pharmpy/tools/amd/run.py
rikardn, line 69: Perhaps use `Literal` in type hint

src/pharmpy/tools/amd/run.py
```diff
@@ -118,6 +119,9 @@ def run_amd(
     mechanistic_covariates : list
         List of covariates to run in a separate proioritized covsearch run. The effects are extracted
         from the search space for covsearch
+    perform_retries: str
+        Weither or not to run retries tool. Valid options are 'skip', 'all_final' or 'amd_final'.
+        Default is 'amd_final'.
```
rikardn, line 124: I think that what you call `amd_final` should be named `final`. Then the same option can be reused for other tools. `all_final` should perhaps be named `subtool_final` or `selected` or something along those lines.

rikardn (changes requested):
I have some comments

src/pharmpy/tools/retries/tool.py
JohanWBorg: I see your point. As range and interval (for me at least) represent the possible numbers rather than the percentage of allowed increase/decrease, I decided to change it to `degree`. I had a hard time coming up with other alternatives, but please let me know if some other word comes to mind.

src/pharmpy/tools/amd/run.py
JohanWBorg: Changed to `retries_strategy`

JohanWBorg:
@rikardn Comments should be resolved now

src/pharmpy/tools/retries/tool.py
rikardn: I think degree is fine. I couldn't find a better word either.

docs/amd.rst
```diff
@@ -454,6 +455,26 @@ For an entire AMD run, it is possible to get a maximum of three covsearch runs,
 | Exploratory         | The remaining covariates are tested after all mechanistic covariates have been tested.  |
 +---------------------+-----------------------------------------------------------------------------------------+
 
+Retries
+~~~~~~~~~~
+
+If ``retries_strategy`` is set to 'all_final', the retries tool will be run on the final model from each subtool.
```
rikardn, line 461: Does this reflect the changes in values for `retries_strategy` ?

src/pharmpy/tools/amd/run.py
```diff
@@ -66,6 +66,7 @@ def run_amd(
     strictness: Optional[str] = "minimization_successful or (rounding_errors and sigdigs>=0.1)",
     dv_types: Optional[dict] = None,
     mechanistic_covariates: Optional[List[str]] = None,
+    retries_strategy: Optional[Literal["final", "all_final", "skip"]] = "final",
```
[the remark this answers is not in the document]
JohanWBorg: Ah, thanks, will correct this

PR #2420, opened 2023-12-06
Integrate parameter_sampling in retries tool

JohanWBorg:
- parameter_sampling can now sample on both normal and UCP scale.
- Retries tool using create_rng instead of random.random
- Added argument "rng" to AMD tool for random sampling (currently only for retries tool)

### DIFF
--- a/docs/retries.rst
+++ b/docs/retries.rst
@@ -46,7 +46,7 @@ The arguments of the retries tool are listed below.
 +-------------------------------------------------+---------------------------------------------------------------------+
 | number_of_candidates                            | Number of retry-models to run. 5 is used as default                 |
 +-------------------------------------------------+---------------------------------------------------------------------+
-| degree                                          | Determine the allowed increase/decrease for the randomly generated  |
+| fraction                                          | Determine the allowed increase/decrease for the randomly generated|
 |                                                 | new initial estimates. Default is 0.1 (10%).                        |
 +-------------------------------------------------+---------------------------------------------------------------------+
 | :ref:`scale<scales_retries>`                    | Scale to use when randomizing the initial estimates. Currently      |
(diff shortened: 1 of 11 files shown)


src/pharmpy/modeling/parameter_sampling.py
```diff
@@ -85,6 +85,9 @@ def _sample_from_function(
     """
     rng = create_rng(rng)
 
+    parameter_estimates = parameter_estimates[
```
rikardn, line 88: There is a better way of doing this. Use `model.parameters.fixed` eller `model.parameters.nonfixed

src/pharmpy/modeling/parameter_sampling.py
```diff
@@ -181,8 +184,12 @@ def sample_parameters_uniformly(
...
         samples = np.empty((n, len(lower)))
         for i, (x, a, b) in enumerate(zip(pe, lower, upper)):
-            lower = max(a, x - x * fraction)
-            upper = min(b, x + x * fraction)
+            if scale == 'normal':
+                lower = max(a, x - x * fraction)
+                upper = min(b, x + x * fraction)
+            elif scale == 'UCP':
+                lower = 0.1 - 0.1 * fraction
+                upper = 0.1 + 0.1 * fraction
```
rikardn, line 192: For sampling on UCP scale don't we still want the resulting parameter samples on the normal scale?

rikardn: Or perhaps it is faster to do conversion later.

docs/retries.rst
```diff
@@ -55,6 +55,8 @@ The arguments of the retries tool are listed below.
 | prefix_name                                     | String determining prefix of model names such that models are named |
 |                                                 | {prefix_name}_retries_run2 for instance.                            |
 +-------------------------------------------------+---------------------------------------------------------------------+
+| rng                                             | A random number generator or seed to use for random sampling.
```
rikardn, line 58: I see that we sometimes call this `rng` and sometimes `seed`. Actually we support either or. What do you think we should call it to be consistent?

rikardn (changes requested):

src/pharmpy/modeling/parameter_sampling.py
JohanWBorg: updated to use model.parameters.nonfixed instead

PR #2462, opened 2023-12-08
Remove overuse of Pharmpy in user guide titles

JohanWBorg:


docs/cli.rst
```diff
@@ -1,5 +1,5 @@
 ==================================
```
rikardn, line 1: Let header fit in `===`

rikardn (changes requested):

rikardn:
Now LGTM!

PR #2484, opened 2023-12-13
ModelEntry in modelfit

JohanWBorg:


src/pharmpy/tools/amd/run.py
```diff
@@ -396,9 +396,8 @@ def run_amd(
         model = filter_dataset(model, 'DVID != 2')
 
     if results is None:
-        model = run_tool('modelfit', model, path=db.path / 'modelfit', resume=resume)
-    else:
-        model = model.replace(modelfit_results=results)
+        results = run_tool('modelfit', model, path=db.path / 'modelfit', resume=resume)
+    model = model.replace(modelfit_results=results)
```
rikardn, line 400: The idea is that `Model` should not contain `modelfit_results`

src/pharmpy/tools/external/nlmixr/run.py
```diff
@@ -21,32 +21,15 @@
...
-
-    Returns
-    -------
-    model : pharmpy.model.Model
-        Model with accompanied results.
+def execute_model(model_entry, db, evaluate=False, path=None):
+    assert isinstance(model_entry, ModelEntry)
+    model = model_entry.model
 
-    """
```
rikardn, line 49: Was the docstring removed on purpose?

src/pharmpy/tools/external/nonmem/run.py
```diff
@@ -17,16 +17,10 @@
...
-    if isinstance(model_or_model_entry, ModelEntry):
-        model = model_or_model_entry.model
-
-        # TODO : Remove after complete integration of ModelEntry
-        if model.modelfit_results is not None:
-            model = model.replace(modelfit_results=None)
-            model_or_model_entry = ModelEntry.create(model, parent=model_or_model_entry.parent)
-    else:
-        model = model_or_model_entry
+def execute_model(model_entry, db):
```
rikardn, line 20: Why can't we have model here? And create the `ModelEntry` in the function.

src/pharmpy/tools/external/rxode/run.py
```diff
@@ -14,27 +14,14 @@
...
-        An pharmpy model object.
-    db : database for storing model results
-        Name of folder in home directory to store resulting files in.
-
-    Returns
-    -------
-    model : pharmpy.model.Model
-        Model with accompanied results.
-
-    """
```
rikardn, line 37: Lost docstring. Perhaps on purpose?

src/pharmpy/tools/modelfit/tool.py
```diff
@@ -20,7 +20,7 @@ def create_workflow(
     n : int
         Number of models to fit. This is only used if the tool is going to be combined with other tools.
     tool : str
-        Which tool to use for fitting. Currently, 'nonmem' or 'nlmixr' can be used.
+        Which tool to use for fitting. Currently, 'nonmem', 'nlmixr', 'rxode' can be used.
```
rikardn, line 23: This seems to indicate that rxode can fit, but it can really only evaluate (predict).

src/pharmpy/tools/modelfit/tool.py
```diff
@@ -3,7 +3,7 @@
 from pharmpy.model import Model
 from pharmpy.workflows import ModelEntry, Task, Workflow, WorkflowBuilder
 
-SupportedPlugin = Literal['nonmem', 'nlmixr']
+SupportedPlugin = Literal['nonmem', 'nlmixr', 'rxode']
```
rikardn, line 6: Perhaps it should be called `SupportedPlugins` now, although `Plugin` should perhaps be something else. `SupportedExternalTools` ?

tests/integration/test_fit.py
```diff
@@ -109,17 +109,6 @@ def test_execute_model_nonmem(tmp_path, testdata):
         model = Model.parse_model('pheno_real.mod')
         model = model.replace(datainfo=model.datainfo.replace(path=tmp_path / 'pheno.dta'))
 
-        db = LocalDirectoryToolDatabase('db_model')
-
-        model = execute_model(model, db)
```
rikardn, line 114: Do we test `execute_model` with `db` elsewhere?

rikardn (changes requested):

src/pharmpy/tools/amd/run.py
stellabelin: I think this is temporary until `ModelEntry` is implemented in AMD? We already did this before this PR.

src/pharmpy/tools/external/nonmem/run.py
stellabelin: The `parent` attribute is set when creating the candidate models in tools, so in order to not have to attach that information after fitting it takes a `ModelEntry`.

stellabelin (changes requested):
An additional comment

src/pharmpy/tools/external/nlmixr/run.py
JohanWBorg: It was removed on purpose as the corresponding function in NONMEM has no docstring and the function is not intended to be used by a user. The same was done for the docstring in `pharmpy.tools.external.rxode/run.py`

src/pharmpy/tools/external/rxode/run.py
JohanWBorg: See comment above

src/pharmpy/tools/modelfit/tool.py
JohanWBorg: This was done in order to be able to run rxode models in the same way as the for the other tools (instead of having to use execute model and define your own database).
Should I remove rxode support from this function?

src/pharmpy/tools/amd/run.py
rikardn: Ok, so AMD is still on todo.

src/pharmpy/tools/external/nlmixr/run.py
rikardn: Ok, makes sense. This falls into some future documentation for how to support a new external tool.

tests/integration/test_fit.py
JohanWBorg: Yes, within the same test just further down. The removed parts of the test were for a `Model` input object, which is no longer supported

src/pharmpy/tools/modelfit/tool.py
rikardn: No, I didn't mean that support should be removed. Rather that the docstring should clarify, but it might be difficult to do because the tool is named `Modelfit`, which is starting to become a bit of a misnomer in itself. So keep this for now.

src/pharmpy/tools/modelfit/tool.py
```diff
@@ -34,10 +34,16 @@ def create_workflow(
     >>> from pharmpy.tools import run_modelfit     # doctest: +SKIP
     >>> run_modelfit(model)   # doctest: +SKIP
     """
-
-    wf = create_fit_workflow(model_or_models, n, tool)
+    if model_or_models:
+        if not isinstance(model_or_models, Iterable):
+            model_or_models = [model_or_models]
+        modelentries = [
+            ModelEntry.create(model=model, modelfit_results=model.modelfit_results)
```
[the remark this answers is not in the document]
JohanWBorg: You are right! Will remove it.

src/pharmpy/tools/amd/run.py
stellabelin: Yes, it's the final tool to be implemented.

src/pharmpy/tools/external/nonmem/run.py
JohanWBorg: I guess we could still have support for a `Model` object but it seems a bit counteractive to the `ModelEntry` implementation

rikardn: I agree now to keep this. My thinking was that the `execute_model` was the source of the `ModelEntry` objects. But  `parent` and potentially future other things cannot be reconciled with this point of view.

tests/integration/test_fit.py
rikardn: Ok, great!

rikardn:
Now this LGTM

PR #2486, opened 2023-12-13
Move retries integration tests to unit tests

JohanWBorg:


rikardn:
Conflicts after I changed from having a `start_model` with `modelfit_results` attached to a tuple of `model` and `results`

PR #2493, opened 2023-12-14
Remove unneccessary use of model.modelfit_results in COVsearch

JohanWBorg:


rikardn:
Thanks for fixing this!

PR #2574, opened 2024-01-10
Change naming of strategies and add SIR, SRI, RSI

JohanWBorg:
Change naming of
SIRIAC - all
SIRIACIR - retries
and add three new methods
SIR, SRI and RSI

Also made updates to the documentation for AMD to make it a bit clearer on how the different modeltypes affect the different strategies. Comments are greatly appreciated :)

### DIFF
--- a/src/pharmpy/tools/amd/run.py
+++ b/src/pharmpy/tools/amd/run.py
@@ -48,7 +48,7 @@ from pharmpy.workflows.results import ModelfitResults
 from ..run import run_tool
 from .results import AMDResults
 
-ALLOWED_STRATEGY = ["SIRIAC", "SIRIACIR"]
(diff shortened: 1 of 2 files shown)


src/pharmpy/tools/amd/run.py
```diff
@@ -1038,7 +1044,7 @@ def validate_input(
     results: Optional[ModelfitResults] = None,
     modeltype: str = 'basic_pk',
     administration: str = 'oral',
-    strategy: Optional[str] = "SIRIAC",
+    strategy: Optional[str] = "all",
```
rikardn, line 1047: This is not `Optional`, i.e. `None` is not allowed.

rikardn (changes requested):

rikardn:
LGTM

PR #2632, opened 2024-01-22
Fix brute_force_block_structure algorithm check

JohanWBorg:


src/pharmpy/tools/iivsearch/tool.py
```diff
@@ -155,7 +155,18 @@ def start(
 
     sum_models = [summarize_modelfit_results(input_res)]
 
-    for algorithm_cur in list_of_algorithms:
+    for algorithm_cur in list_of_algorithms.copy():
```
rikardn, line 158: Do not copy here (see next comment)

src/pharmpy/tools/iivsearch/tool.py
```diff
@@ -155,7 +155,18 @@ def start(
...
+            algorithm_cur == 'brute_force_block_structure'
+            and len(
+                set(base_model_entry.model.random_variables.iiv.names).difference(
+                    _get_fixed_etas(base_model_entry.model)
+                )
+            )
+            <= 1
+        ):
+            list_of_algorithms.remove('brute_force_block_structure')
+            break
```
rikardn, line 169: Replace with `continue`

rikardn (changes requested):

src/pharmpy/tools/iivsearch/tool.py
JohanWBorg: Fixed it by introducing new variable for the applied algorithms

rikardn (approved):
LGTM!

PR #2695, opened 2024-02-20
Bottom-Up IIVSearch

JohanWBorg:


src/pharmpy/tools/iivsearch/tool.py
```diff
@@ -27,7 +33,15 @@
...
     ('no_add', 'add_diagonal', 'fullblock', 'pd_add_diagonal', 'pd_fullblock')
 )
-IIV_ALGORITHMS = frozenset(('brute_force', 'brute_force_no_of_etas', 'brute_force_block_structure'))
+IIV_ALGORITHMS = frozenset(
+    (
+        'td_brute_force',
+        'bu_brute_force',
+        'td_exhaustive_no_of_etas',
+        'bu_exhaustive_no_of_etas',
+        'td_exhaustive_block_structure',
```
rikardn, line 42: We need more eyes on the new naming. @stellabelin could you comment?

rikardn (changes requested):

src/pharmpy/tools/iivsearch/tool.py
stellabelin: I think the current naming is a bit confusing, but not sure what naming would make things clearer. Some thoughts:

- I think that it's a bit confusing that we both have `brute_force` and `exhaustive` in the naming, to me it's not apparent what the difference between them are from the name. 
- Bottom up is not a true exhaustive right? Since it stops when no better model is found? Perhaps `stepwise` would be useful here? 
- In general I find top down/bottom up a bit confusing here, given that `bu_brute_force` still uses the top bottom approach for the block structure (Idk, maybe in the future we'll have a bottom up approach for that too).
- The abbreviation `td` and `bu` is a bit confusing, however using the full name would render the names of these algorithms very long. 

Not sure what a good alternative to naming would be, I was otherwise thinking that we could add an option, so that `algorithm: Literal['top_down_exhaustive', 'bottom_up_exhaustive']` and `eta_selection: Literal['variance', 'covariance', 'both']`, however more options is also more complex so not sure if this is a good idea.

src/pharmpy/modeling/parameters.py
```diff
@@ -169,11 +169,16 @@ def update_inits(
     {'PTVCL': 0.00469555, 'PTVV': 0.984258, 'THETA_3': 0.15892, 'IVCL': 0.0293508, 'IVV': 0.027906, ...}
 
     """
+    all_parameter_estimates = model.parameters.inits
+    all_parameter_estimates.update(parameter_estimates)
+
     # FIXME: Can be combined with set_initial_estimates
     if move_est_close_to_bounds:
-        parameter_estimates = _move_est_close_to_bounds(model, parameter_estimates)
+        all_parameter_estimates = _move_est_close_to_bounds(model, all_parameter_estimates)
```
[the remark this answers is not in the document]
JohanWBorg: You are right, that was an unnecessary precaution on my end.

src/pharmpy/tools/iivsearch/tool.py
JohanWBorg: New naming has now been added using two arguments.
One main `algorithm` argument takes the values of `top_down_exhaustive` and `bottom_up_stepwise`. If possible, the same algorithm will be used in both the search for the number of etas and for the covariance structure. However, another option will be used if the chosen algorithm only exists for one of the two searches.

Then there also exists a second optional argument called `correlation_algorithm` which can define if a separate algorithm is to be used for the covariance structure search. (However, this search currently only supports `top_down_exhaustive`)

src/pharmpy/tools/iivsearch/tool.py
```diff
@@ -324,6 +382,39 @@ def post_process(rank_type, cutoff, strictness, input_model_entry, base_model_na
         strictness=strictness,
     )
 
+    if rank_type == "mbic":
```
stellabelin, line 385: How come this is needed here? Is it not possible to just set `n_predicted`, `n_expected` etc if it's mBIC, `None` otherwise? Was thinking so we don't need to double maintain.

src/pharmpy/workflows/optimize.py
```diff
@@ -14,12 +14,14 @@ def _scatter_computation(Future, client, computation):
     #  - Any other value like 1, to be interpreted literally
     #  - A task like (inc, 'x') (see below)
     #  - A list of computations, like [1, 'x', (inc, 'x')]
-
     if isinstance(computation, tuple):
-        return (
-            computation[0],
-            *map(lambda c: _scatter_computation(Future, client, c), computation[1:]),
-        )
+        if len(computation) == 0:
```
stellabelin, line 18: I think it could be helpful to write a comment here of why it's needed.

tests/integration/test_iivsearch.py
```diff
@@ -137,23 +162,36 @@ def test_brute_force(tmp_path, model_count, start_modelres):
...
+    ('algorithm', 'correlation_algorithm', 'iiv_strategy', 'no_of_candidate_models'),
+    (
+        ('top_down_exhaustive', 'skip', 'add_diagonal', 15),
+        ('top_down_exhaustive', 'skip', 'fullblock', 15),
+        ('bottom_up_stepwise', 'skip', 'add_diagonal', 6),
+        ('bottom_up_stepwise', 'skip', 'fullblock', 6),
+    ),
 )
-def test_no_of_etas_iiv_strategies(tmp_path, model_count, start_modelres, iiv_strategy):
+def test_no_of_etas_iiv_strategies(
```
stellabelin, line 173: Is it needed to add both the tests for bottom up? Was thinking since these are quite slow. Perhaps enough with `add_diagonal`?

src/pharmpy/tools/iivsearch/algorithms.py
```diff
@@ -42,8 +52,132 @@ def brute_force_no_of_etas(base_model, index_offset=0, keep=None):
...
+        "initialize_stepwise_BU",
+        initialize_stepwise_BU,
+        base_model,
+        index_offset,
+    )
+    wb.add_task(stepwise_task)
+    return wb
+
+
+def initialize_stepwise_BU(context, base_model, index_offset, base_model_entry):
```
stellabelin, line 67: Is this an initialization? It's the full algorithm right?

stellabelin (commented):
Looks good, some minor comments on the code.

src/pharmpy/modeling/parameters.py
```diff
@@ -185,10 +186,12 @@ def _get_nonfixed_rvs(model):
...
 
 
-def _move_est_close_to_bounds(model: Model, pe):
+def _move_est_close_to_bounds(model: Model, est):
     rvs, pset = _get_nonfixed_rvs(model), model.parameters
-    est = pe.to_dict()
-    sdcorr = rvs.parameters_sdcorr(est)
+
+    all_parameter_estimates = model.parameters.inits
+    all_parameter_estimates.update(est)  # Need all numerical values for sdcorr
```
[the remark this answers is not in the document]
JohanWBorg: Updated name as per your suggestion!

tests/integration/test_iivsearch.py
JohanWBorg: I can remove the `fullblock` for the bottom up approach but then that will not have coverage. Should both arguments still be tested for the top down approach?

src/pharmpy/tools/iivsearch/algorithms.py
JohanWBorg: renamed `stepwise_BU_algorithm`

src/pharmpy/tools/iivsearch/tool.py
JohanWBorg: Moved the option to have it a bit more straightforward. However, the n_predictied and n_expected differ if running bottom up vs running top down. The bottom up approach uses a value different from the default which is the reason for it being different.

tests/integration/test_iivsearch.py
JohanWBorg: Kept both for now. Please let me know if you wish to have one removed

stellabelin: I think when dry-run is implemented we can move them for the coverage. Do you know how long the added tests take currently?

docs/iivsearch.rst
```diff
@@ -94,78 +100,44 @@ Algorithms
...
+| ``'bottom_up_stepwise'``            | Iteratively adds all available IIV, one at a time. After each addition, the    |
+|                                     | best model is selected and another ETA. The algorithm stops when no better     |
+|                                     | is found when adding a new ETA. The following covariance structure search uses |
+|                                     | the same approach as 'top_down_exhaustive'                                     |
+|                                     | Currently not supported for 'correlation_algorithm'
++-------------------------------------+--------------------------------------------------------------------------------+
+| ``'skip'``                          | Set this argument if you are certain to skip either the search for number of   |
+|                                     | etas (algorithm='skip') or to skip the serach for the best covariance structure|
+|                                     | (correlation_algorithm='skip'). However, if algorithm is set to skip, then     |
+|                                     | correlation_algorithm need to be set to a valid value.                         |
```
stellabelin, line 124: Maybe use '' on correlation_algorithm to be consistent.

docs/iivsearch.rst
```diff
@@ -203,7 +175,50 @@ algorithm for the block structure, by first choosing the number of etas then the
             s7 -> s12
 
         }
+        
+Bottom up stepwise search
+-------------------------
+
+The ``bottom_up_stepwise`` algorithm differ slightly from the ``top_down_exhaustive`` as the models are created
```
stellabelin, line 182: Picky but it differs quite a bit in terms of algorithm no?

docs/iivsearch.rst
```diff
@@ -203,7 +175,50 @@ algorithm for the block structure, by first choosing the number of etas then the
...
+in iterative steps, each adding a single ETA. The algorithm will create a model with all possible IIVs and in the first step
+it will remove all but one. This ETA will be on clearance (CL) if possible. If not, the first parameter in alphabetical order
+will have an ETA. This model is then run and its results are used to update the initial estimates of the model. In the next step,
+a candidate model is created for each remaining parameter that could have an ETA put on it. All models are run, and the best model
+is chosen for the next step, updating the initial values once more.
+
+If no better model can be found, the algorithm stops.
```
stellabelin, line 189: Maybe specify what we mean by better, that we use the selection criteria specified by the user (since it could also mean using LRT or something).

docs/iivsearch.rst
```diff
@@ -203,7 +175,50 @@ algorithm for the block structure, by first choosing the number of etas then the
...
 
+However, this algorithm is not possible to run for the covariance structure search and thusly it will use the algorithm ``top_down_exhaustive``
```
stellabelin, line 191: Maybe "is not supported" rather than "not possible", later a stepwise approach might be possible right?

stellabelin, line 191: Maybe "is not supported" rather than "not possible", later a stepwise approach might be possible

docs/iivsearch.rst
```diff
@@ -94,78 +100,44 @@ Algorithms
...
+    We recommend setting both arguments if specific algorithms are wanted.
+
++-------------------------------------+--------------------------------------------------------------------------------+
+| Algorithm                           | Description                                                                    |
++=====================================+================================================================================+
+| ``'top_down_exhaustive'``           | Removes available IIV in all possible combinations. The covariance structure   |
+|                                     | search will search all possible IIV variance and covariance structure          |
++-------------------------------------+--------------------------------------------------------------------------------+
+| ``'bottom_up_stepwise'``            | Iteratively adds all available IIV, one at a time. After each addition, the    |
+|                                     | best model is selected and another ETA. The algorithm stops when no better     |
```
stellabelin, line 116: "when no better model", to me the sentence sounds a bit incomplete otherwise.

stellabelin (commented):
And here are some documentation comments! The changes you made to the code LGTM!

stellabelin (approved):
LGTM!

PR #2715, opened 2024-02-27
Simplify covariate extraction and allow param/covariate combination in AMD

JohanWBorg:
Fixes crashing AMD integration test
Simplifies covariate effect extraction
Allows for param/covariate combinations (as tuples) for structural covariate effects in AMD

### DIFF
--- a/src/pharmpy/modeling/covariate_effect.py
+++ b/src/pharmpy/modeling/covariate_effect.py
@@ -29,7 +29,7 @@ EffectType = Union[Literal['lin', 'cat', 'piece_lin', 'exp', 'pow'], str]
 OperationType = Literal['*', '+']
 
 
-def get_covariates(model: Model) -> dict[list]:
+def get_covariate_effects(model: Model) -> dict[list]:
     """Return a dictionary of all used covariates within a model
 
     The dictionary will have parameter name as key with a connected value as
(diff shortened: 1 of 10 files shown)


src/pharmpy/tools/amd/run.py
```diff
@@ -913,63 +913,38 @@ def _run_mechanistic_exploratory_covariates(model, modelfit_results):
...
                     results=modelfit_results,
                     path=path / 'covsearch_mechanistic',
                 )
-                index_offset = int(
-                    res.summary_tool.iloc[-1].name[-1][13:]
-                )  # Get largest number of run
+                model_db = res.tool_database.model_database
+                all_models = [model_db.retrieve_model(model) for model in model_db.list_models()]
+                covsearch_model_number = [
+                    int(model.name[13:])
```
stellabelin, line 942: I'm guessing the 13 here is the length of `covsearch_run`, I see this both as a bit confusing (it's not apparent what 13 comes from) and as a bit risky if we e.g. change the names of the candidates. Perhaps regex can be used instead to capture the number? This would maybe address the first concern and at least make the second concern explicit.

src/pharmpy/tools/amd/run.py
```diff
@@ -999,6 +974,47 @@ def _run_mechanistic_exploratory_covariates(model, modelfit_results):
...
+        else:
+            current_cov = []
+            current_param = []
+            for cov in cov_statement.covariate:
+                if cov in mechanistic_covariates:
+                    current_cov.append(cov)
+                    current_param.append(cov_statement.parameter)
+                else:
+                    for param in cov_statement.parameter:
+                        if {cov, param} in mechanistic_covariates:
```
stellabelin, line 996: Maybe misunderstanding but if the set-conversion happens at line 978 why is it needed here too? Could line 978 be removed and just this kept?

src/pharmpy/tools/amd/run.py
```diff
@@ -1118,6 +1134,51 @@ def validate_input(
...
+                cov_found = False
+                par_found = False
+                for a in c:
+                    if a in allowed_covariates:
+                        if cov_found:
+                            raise ValueError(
+                                f'`mechanistic_covariates` contain invalid argument: got `{c}`,'
+                                f' which contain two covariates.'
+                                f' Tuples need to be given with one parameter and one covariate.'
+                            )
```
stellabelin, line 1164: Since this error is identical for all three if-statements, perhaps it can be stored as a variable.

src/pharmpy/tools/amd/run.py
```diff
@@ -1118,6 +1134,51 @@ def validate_input(
...
+                        par_found = True
+                    else:
+                        raise ValueError(
+                            f'`mechanistic_covariates` contain invalid argument: got `{c}`,'
+                            f' which contain {a}.'
+                            f' Tuples need to be given with one parameter and one covariate.'
+                        )
+            # Typechecking performed elsewhere?
```
stellabelin, line 1180: This comment should be removed right? And to answer the question, this function does the type checking:
https://github.com/pharmpy/pharmpy/blob/cb5a277c678a0a24cff2bc7586e9f9109235ad18/src/pharmpy/internals/fn/type.py#L20
It is called using a decorator on the `validate_input` function of each tool, e.g.:
https://github.com/pharmpy/pharmpy/blob/cb5a277c678a0a24cff2bc7586e9f9109235ad18/src/pharmpy/tools/modelsearch/tool.py#L294-L296

src/pharmpy/tools/covsearch/tool.py
```diff
@@ -270,16 +273,17 @@ def func_description(effect_funcs, model=None, add=True):
...
+        potential_extension = func_description(optional_remove, filtered_model, add=True)
+        for _, optional_func in optional_remove.items():
             filtered_model = optional_func(filtered_model)
         if len(potential_extension) != 0:
             if not description:
                 description.append("REMOVED")
             description.extend(potential_extension)
 
     # Add all mandatory covariates
+    print(mandatory_funcs)
```
stellabelin, line 286: This should be removed right?

stellabelin (commented):
Looks good, some comments!

src/pharmpy/modeling/covariate_effect.py
```diff
@@ -47,6 +47,7 @@ def get_covariates(model: Model) -> dict[list]:
     """
     parameters = get_individual_parameters(model)
     covariates = get_model_covariates(model)
+    covariates = [c for c in covariates if model.datainfo[str(c)].type in ["covariate", "unknown"]]
```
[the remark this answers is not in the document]
JohanWBorg: Hmm, yes perhpas. It is defined there as having 
```
a data item affecting the model prediction excluding dosing items that 
are not used in model code.
```
Maybe this extra filtering on "covariate" and "unknown" is an unnecessary precaution?

src/pharmpy/modeling/covariate_effect.py
```diff
@@ -58,79 +59,70 @@ def get_covariates(model: Model) -> dict[list]:
     res = defaultdict(list)
     for param, covariates in param_w_cov.items():
         for cov in covariates:
-            eff, op = get_covariate_effect(model, param, str(cov))
-            res[(param, cov)].append((eff, op))
-
+            effop = _get_covariate_effect(model, param, cov)
```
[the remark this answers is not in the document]
JohanWBorg: Changed it to be a bit clearer :)

src/pharmpy/tools/amd/run.py
JohanWBorg: Will use regex to find the number in the end of the name instead. Thanks for your input!

src/pharmpy/tools/amd/run.py
JohanWBorg: They are not identical but very similar since they are all about how the tuples are to be given.
They all start with `mechanistic_covariates contain invalid argument: got {c},` and end with the hint `Tuples need to be given with one parameter and one covariate.`

but the second line gives information on what is actually wrong.

- `which contain two covariates.`
- `' which contain two parameters.'`
- `' which contain {a}.'` (where a is the argument causing the error)

I can change it to store the first/last line as two separate variables if anything but that might be messy in its own way

src/pharmpy/tools/amd/run.py
stellabelin: Ah okay, I understand now, makes sense!

src/pharmpy/tools/amd/run.py
stellabelin: My bad, then this is better!

src/pharmpy/modeling/covariate_effect.py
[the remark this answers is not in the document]
JohanWBorg: Removed it for now to use the same definition in both cases

stellabelin:
LGTM!

PR #2786, opened 2024-03-14
Default keep=["CL"] in IIVSearch

### LINKED ISSUE #2776: Have keep=CL as default also for iivsearch when run stand alone
This is a source of confusion for users that the default is different in the tool and in amd. We should have the same default.

JohanWBorg:
Should fix #2776.
Warns the user if parameters are given in the "keep" argument that are not found in the model (this includes the default value of "CL")

### DIFF
--- a/src/pharmpy/tools/iivsearch/tool.py
+++ b/src/pharmpy/tools/iivsearch/tool.py
@@ -49,7 +49,7 @@ def create_workflow(
     cutoff: Optional[Union[float, int]] = None,
     results: Optional[ModelfitResults] = None,
     model: Optional[Model] = None,
-    keep: Optional[List[str]] = None,
+    keep: Optional[List[str]] = ["CL"],
     strictness: Optional[str] = "minimization_successful or (rounding_errors and sigdigs>=0.1)",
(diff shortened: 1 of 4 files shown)


src/pharmpy/tools/iivsearch/tool.py
```diff
@@ -423,12 +423,12 @@ def flatten_list(lst):
...
     algorithm, iiv_strategy, rank_type, model, keep, strictness, correlation_algorithm
 ):
-    if keep:
+    if keep and model:
         for parameter in keep:
             try:
                 has_random_effect(model, parameter, "iiv")
             except KeyError:
-                raise ValueError(f"Parameter {parameter} has no iiv.")
+                warnings.warn(f"Parameter {keep} has no iiv and is ignored")
```
stellabelin, line 431: Why is this a warning now?

JohanWBorg: As "CL" might not always be part of the model (but is now the default value) this needed to be changed. Otherwise we might raise with the default values

stellabelin:
LGTM!

Issue #2780, opened 2024-03-13
Use get_individual_parameters instead of get_rv_parameters


JohanWBorg:
`get_individual_parameters` has implemented new logic for extracting parameters which avoids problems present in `get_rv_parameters`.

An example of this can be seen below.

The ODE system (in K-rate terms) for a PSC drug metabolite could be defined as :
```
K12 = (1 - FPRE)/MAT
K13 = FPRE/MAT
K23 = CL/V
K30 = CLM/VM
```
Where 1 : DEPOT, 2 : CENTRAL and 3 : METABOLITE

The ETAs connected to FPRE and MAT are not connected to these parameters but are instead wrongfully connected to K12. This due to the fact that this expression is not classified as "synthetic" in the logic applied to `get_rv_parameters`

By instead relying on `get_individual_parameters` when extracting which parameter an ETA is connected to, this problem can be avoided.

PR #2809, opened 2024-03-21
Implement get_individual_parameters within get_rv_parameters

JohanWBorg:
Should solve #2780. `get_rv_parameters` remains but will only be able to find parameters that are also part of the results from `get_individual_parameters`. If none of these parameters are connected to the given IIV, an error is raised. This is different from its past behavior where it previously returned an empty list in the case where to connected parameter could not be found.


stellabelin (approved):
LGTM!

PR #2811, opened 2024-03-21
Add method ModelFeatures.replace_features

stellabelin:


src/pharmpy/tools/mfl/parse.py
```diff
@@ -309,6 +315,38 @@ def replace(self, **kwargs):
...
+            EffectComp: 'effect_comp',
+            IndirectEffect: 'indirect_effect',
+            Metabolite: 'metabolite',
+        }
+        mfl_list = _parse(mfl_str)
+        kwargs = {}
+        for statement in mfl_list:
+            key = key_dict[type(statement)]
+            if type(statement) in (Transits, Peripherals, Covariate, IndirectEffect):
+                no_of_modes = len(statement.counts)
```
JohanWBorg, line 336: I believe the `Covariate` statements do not have the `.counts` attribute. If the number of covariate effects it represents wants to be used, I would check the combinations of parameter/covariate

src/pharmpy/tools/mfl/parse.py
```diff
@@ -309,6 +315,38 @@ def replace(self, **kwargs):
...
+                value = (statement,)
+            else:
+                no_of_modes = len(statement.modes)
+                value = statement
+            if no_of_modes > 1:
```
JohanWBorg, line 341: I am not entirely sure I follow why you would not be allowed to replace an attribute with a feature with more than one "mode". For instance `ABSORPTION([FO,ZO])`. Is there a specific reason as to why this should not be allowed? As it is supported in the class.

tests/tools/test_mfl.py
```diff
@@ -1406,6 +1406,27 @@ def test_mfl_function_filtration(load_model_for_test, pheno_path):
...
+    ss = "PERIPHERALS(2);PERIPHERALS(1, MET)"
+    mfl_new = mfl_original.replace_features(ss)
+    assert mfl_new.peripherals == (Peripherals((2,)), Peripherals((1,), (Name('MET'),)))
+
+    with pytest.raises(ValueError):
+        ss = "ABSORPTION([FO,ZO])"
+        mfl_original.replace_features(ss)
+
+    with pytest.raises(ValueError):
+        ss = "ABSORPTION([FO,ZO])"
```
JohanWBorg, line 1426: Same question as above

src/pharmpy/tools/mfl/parse.py
```diff
@@ -309,6 +315,38 @@ def replace(self, **kwargs):
...
+                raise ValueError(f'Cannot set more than one mode for type: `{key}`')
+            if key_dict[Peripherals] in kwargs.keys():
+                kwargs[key] += value
+            else:
+                kwargs[key] = value
```
JohanWBorg, line 346: As for instance `Covariate` effects can be given in more than one statement, wouldn't this overwrite the previously stored value if multiple statements are given?
This goes hand in hand with my question above regarding if this should be supported

JohanWBorg (changes requested):
Looks good, some minor questions :)

src/pharmpy/tools/mfl/parse.py
stellabelin: Fair point, originally I came from the side of the modelbuilder where it's always one feature at the time, but I don't think it needs to be restricted to that. Will change!

src/pharmpy/tools/mfl/parse.py
stellabelin: `Covariate` should be included, an oversight. Will change!

JohanWBorg:
@stellabelin Nice one! LGTM now :)

PR #2826, opened 2024-03-26
Expand logic for covariate effect extraction

JohanWBorg:
- Removal of the automatic addition of "ELSE" statements in if statements where it is not explicitely stated
- Correct expansion of exponential covariate effects when combined with other exponential terms
- More test cases for get_covariate_effects


rikardn (approved):
Unfortunate that our example model has a recursive variable. Anyway: LGTM

PR #2868, opened 2024-04-08
Correct extraction of redundant K-rate individual parameters

JohanWBorg:
A model being created with first order absorption as well as a number of transit compartments will have somewhat redundant k-rate parameters. The following example would create such a model

```
model = load_example_model("pheno")
model = set_first_order_absorption(model)
model = set_transit_compartments(model, 3)
```

The assignments in question would be the following

```
MAT = POP_MAT
KA = 1/MAT
K45 = KA
```

And through this, MAT was previously not caught as an individual parameter.

### DIFF
--- a/src/pharmpy/modeling/expressions.py
+++ b/src/pharmpy/modeling/expressions.py
@@ -768,7 +768,6 @@ def get_individual_parameters(
 
     """
     # FIXME: Support multiple DVs
-
     model = make_declarative(model)
     model = _replace_trivial_redefinitions(model)
     statements = model.statements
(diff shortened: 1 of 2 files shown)


src/pharmpy/modeling/expressions.py
```diff
@@ -978,7 +976,20 @@ def _replace_trivial_redefinitions(model):
...
+        if temp_d[k] in temp_d.keys() and temp_d[k] != k:
+            return update_value(temp_d, d[k])
+        else:
+            return temp_d[k]
+
+    clean_d = d.copy()
+    for k, v in d.items():
+        clean_d[k] = update_value(d, k)
+
+    new = Statements(tuple(keep)).subs(clean_d)
```
rikardn, line 992: Couldn't all of this be done simpler directly in the first for loop directly? 

isn't it simply to apply the subs dict to the rhs for each iteration? Or optionally perform a check similarly to what you did?

In any case it is a good catch!

rikardn (changes requested):

src/pharmpy/modeling/expressions.py
JohanWBorg: Optimized the function to perform the same substitution but within the first for-loop

JohanWBorg:
@rikardn the requested change should now have been solved

PR #2932, opened 2024-04-22
Substitute fixed parameter symbols when calculating new UCP parameters

### LINKED ISSUE #2904: Retries does not work for PKPD models
### Bug report

Get `TypeError: Symbolic matrix cannot be converted to numeric` when trying to run `run_amd` for a PKPD model with default retries strategy.

JohanWBorg:
Should fix #2904

### DIFF
--- a/src/pharmpy/modeling/estimation.py
+++ b/src/pharmpy/modeling/estimation.py
@@ -126,8 +126,16 @@ def calculate_parameters_from_ucp(
     --------
     calculate_ucp_scale : Calculate the scale for conversion from ucps
     """
+    fixed_parameters_dict = model.parameters.fixed.inits
+    for p in model.parameters.names:
+        if p not in fixed_parameters_dict and p not in ucps:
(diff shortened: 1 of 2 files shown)


src/pharmpy/modeling/estimation.py
```diff
@@ -126,8 +126,16 @@ def calculate_parameters_from_ucp(
     --------
     calculate_ucp_scale : Calculate the scale for conversion from ucps
     """
+    fixed_parameters_dict = {
+        p.name: p.init for p in model.parameters if p.fix and p.name not in ucps
+    }
```
rikardn, line 131: Could `model.parameter.fixed.inits` be used to get all fixed values?

rikardn (changes requested):

src/pharmpy/modeling/estimation.py
JohanWBorg: Yes that could be used instead. Changed this as well as added some tests for coverage

rikardn (approved):

PR #2941, opened 2024-04-24
Convert residuals to DataFrame in outlier_count

JohanWBorg:
If only one columns present, this variable turned into a Series


rikardn (approved):
Good catch!

PR #2956, opened 2024-04-26
Restructure derivative handling

JohanWBorg:
Single .derivative attribute instead of splitting ETA/EPS in EstimationStep. Derivatives are always with respect to the prediction.
Derivatives are stored as a tuple of tuples where each tuple represent one derivative. Each element within the tuples are required to be a symbol of type pharmpy.basic.Expr.

### DIFF
--- a/src/pharmpy/model/execution_steps.py
+++ b/src/pharmpy/model/execution_steps.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any, Iterable, Mapping, Optional, Union, overload
 
+from pharmpy.basic import Expr
 from pharmpy.internals.immutable import Immutable, frozenmapping
 
 if TYPE_CHECKING:
@@ -153,8 +154,7 @@ class EstimationStep(ExecutionStep):
         solver_rtol: Optional[int] = None,
         solver_atol: Optional[int] = None,
         tool_options: Optional[frozenmapping[str, Any]] = None,
-        eta_derivatives: Optional[tuple[str, ...]] = None,
-        epsilon_derivatives: Optional[tuple[str, ...]] = None,
+        derivatives: Sequence[Sequence[Expr]] = (),
     ):
         self._method = method
         self._interaction = interaction
@@ -168,8 +168,7 @@ class EstimationStep(ExecutionStep):
         self._keep_every_nth_iter = keep_every_nth_iter
         self._residuals = residuals
         self._predictions = predictions
-        self._eta_derivatives = eta_derivatives
-        self._epsilon_derivatives = epsilon_derivatives
+        self._derivatives = derivatives
         super().__init__(
             solver=solver,
             solver_rtol=solver_rtol,
@@ -196,8 +195,7 @@ class EstimationStep(ExecutionStep):
         solver_rtol: Optional[int] = None,
         solver_atol: Optional[int] = None,
         tool_options: Optional[Mapping[str, Any]] = None,
-        eta_derivatives: Optional[Sequence[str]] = None,
-        epsilon_derivatives: Optional[Sequence[str]] = None,
+        derivatives: Sequence[Sequence[Expr]] = (),
     ):
         method = EstimationStep._canonicalize_and_check_method(method)
         if maximum_evaluations is not None and maximum_evaluations < 1:
@@ -214,6 +212,9 @@ class EstimationStep(ExecutionStep):
             predictions = ()
         else:
             predictions = tuple(predictions)
+        if derivatives:
+            derivatives = EstimationStep._canonicalize_derivatives(derivatives)
+
         if parameter_uncertainty_method is not None:
             parameter_uncertainty_method = parameter_uncertainty_method.upper()
         if not (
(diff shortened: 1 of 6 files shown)


src/pharmpy/model/execution_steps.py
```diff
@@ -153,8 +154,7 @@ def __init__(
         solver_rtol: Optional[int] = None,
         solver_atol: Optional[int] = None,
         tool_options: Optional[frozenmapping[str, Any]] = None,
-        eta_derivatives: Optional[tuple[str, ...]] = None,
-        epsilon_derivatives: Optional[tuple[str, ...]] = None,
+        derivatives: tuple[tuple[Type[Expr]], ...] = (),
```
rikardn, line 157: It should be `Expr` rather than `Type[Expr]`

src/pharmpy/model/execution_steps.py
```diff
@@ -196,8 +195,7 @@ def create(
         solver_rtol: Optional[int] = None,
         solver_atol: Optional[int] = None,
         tool_options: Optional[Mapping[str, Any]] = None,
-        eta_derivatives: Optional[Sequence[str]] = None,
-        epsilon_derivatives: Optional[Sequence[str]] = None,
+        derivatives: tuple[tuple[Type[Expr]], ...] = (),
```
rikardn, line 198: `Expr` rather than `Type[Expr]`

src/pharmpy/model/execution_steps.py
```diff
@@ -272,6 +264,32 @@ def _canonicalize_and_check_method(method: str) -> str:
             )
         return method
 
+    @staticmethod
+    def _canonicalize_derivatives(derivatives: tuple) -> tuple:
+        if isinstance(derivatives, tuple):
```
rikardn, line 269: Hard to follow because of too deep nesting. Check for the problematic case and raise and have no `else`

src/pharmpy/model/execution_steps.py
```diff
@@ -352,14 +370,9 @@ def predictions(self) -> Optional[tuple[str, ...]]:
...
     @property
-    def eta_derivatives(self) -> Optional[tuple[str, ...]]:
-        """List of names of etas for which to calculate derivatives"""
-        return self._eta_derivatives
-
-    @property
-    def epsilon_derivatives(self) -> Optional[tuple[str, ...]]:
-        """List of names of epsilons for which to calculate derivatives"""
-        return self._epsilon_derivatives
+    def derivatives(self) -> tuple[tuple[Type[Expr]], ...]:
```
rikardn, line 373: The same type hint issue

tests/model/test_estimation.py
```diff
@@ -189,13 +193,36 @@ def test_getitem():
 
 
 def test_properties():
-    a = EstimationStep.create('foce', epsilon_derivatives=['EPS(1)'])
-    assert a.epsilon_derivatives == ('EPS(1)',)
-    a = EstimationStep.create('foce', eta_derivatives=['ETA(1)'])
-    assert a.eta_derivatives == ('ETA(1)',)
-    a = EstimationStep.create('foce', predictions=['PRED'])
+    d = tuple(map(Expr.symbol, ('EPS(1)',)))
```
rikardn, line 196: Pharmpy symbols should not contain parentheses. (parsing becomes ambiguous)

rikardn (changes requested):

src/pharmpy/model/execution_steps.py
JohanWBorg: Changed all types to `Expr`

src/pharmpy/model/execution_steps.py
JohanWBorg: Restructured for clarity

tests/model/test_estimation.py
JohanWBorg: Switched the symbols to instead use underlines ("ETA_1")

src/pharmpy/model/execution_steps.py
```diff
@@ -272,6 +264,29 @@ def _canonicalize_and_check_method(method: str) -> str:
             )
         return method
 
+    @staticmethod
+    def _canonicalize_derivatives(derivatives: tuple) -> tuple:
+        if not isinstance(derivatives, tuple):
```
rikardn, line 269: This is not what we want. For `create` any `Sequence` should be fine. We simply convert it to `tuple`.

src/pharmpy/model/execution_steps.py
```diff
@@ -272,6 +264,29 @@ def _canonicalize_and_check_method(method: str) -> str:
...
+            raise ValueError(f"Derivatives must be given in a tuple. Got '{type(derivatives)}")
+
+        if not all(isinstance(d, tuple) for d in derivatives):
+            raise ValueError("Each derivative must be given as 'tuple'")
```
rikardn, line 273: With the same logic as above. Cast and only raise if it cannot be converted.

src/pharmpy/model/execution_steps.py
```diff
@@ -272,6 +264,29 @@ def _canonicalize_and_check_method(method: str) -> str:
...
+        for d in derivatives:
+            for d_arg in d:
+                if not (isinstance(d_arg, Expr) and d_arg.is_symbol()):
+                    raise ValueError(
+                        f"Each derivative argument must be a symbol of type 'Expr'. "
+                        f"Found '{d_arg}' of type {type(d_arg)}"
+                    )
+
+        derivatives = tuple(d if isinstance(d, tuple) else (d,) for d in derivatives)
+        # Sorting of symbols creates unpredictable order
```
rikardn, line 284: Strange comment. The order becomes predictable (at least to us).

rikardn (changes requested):

src/pharmpy/model/execution_steps.py
JohanWBorg: Oh ok, i misunderstood this then, will change it to allow any sequence

src/pharmpy/model/execution_steps.py
JohanWBorg: Understandable. The comment is referring to an issue where sorting "symbols" of `Expr` type gave an unpredictable order. For this reason, strings are used instead.

JohanWBorg: Yes, that would have been a much more accurate description

PR #2960, opened 2024-04-26
Prediction/residual names stored as tuples with certain allowed values

JohanWBorg:


src/pharmpy/model/execution_steps.py
```diff
@@ -206,14 +206,18 @@ def create(
...
-        if residuals is None:
-            residuals = ()
+
+        if isinstance(residuals, tuple):
+            residuals = tuple(sorted(residuals))
         else:
-            residuals = tuple(residuals)
-        if predictions is None:
-            predictions = ()
+            raise ValueError(f"Residuals should be given as a tuple. Recieved {type(residuals)}")
```
rikardn, line 213: Raise only if it cannot be converted to a `tuple`

rikardn (changes requested):

rikardn (approved):

Issue #2900, opened 2024-04-11
Add new parameterizations for categorical covariate effects


rikardn:
The current `CAT` effect sets the mode to 1 (no change) and the others estimate a fractional effect with `1 + THETA`.

We should add `CAT2` having 1 for mode and `THETA` for the others. The thetas would represent increase/decrease. I find it hard to properly name the effect so simply numbering it and documenting which is which should be good enough.

PR #2994, opened 2024-05-06
Alternative categorical covariate effect (CAT2)

JohanWBorg:
New categorical covariate effect (see #2900 for more information)

The initial values of the added theta is set to 0-6 currently to represent the same interval as the "regular" categorical effect 'cat'.


rikardn (approved):
LGTM

PR #2998, opened 2024-05-06
Update extracted derivatives within NONMEM code

JohanWBorg:


rikardn (approved):
Ok, let us keep the part where statements are added. The option to "hide" these statements from Pharmpy is of course there but currently that is not a strategy that we have fully committed to so LGTM

PR #3010, opened 2024-05-08
Update drug metabolite documentation

JohanWBorg:
After previous restructures, the drug metabolite documentation had some descriptions that are no longer accurate. Comments on other changes are more than welcome

### DIFF
--- a/docs/drug_metabolite.rst
+++ b/docs/drug_metabolite.rst
@@ -22,6 +22,10 @@ The code to initiate structsearch for a drug metabolite model in Python/R is sta
                             model=start_model,
                             results=start_model_results)
 
+.. note::                          
+    The tool expect the model to have a dataset with two DVIDs. DVID=1 connected to the parent
+    drug and DVID=2 which is representing the metabolite.
+
 +-------------------------------------------------+-----------------------------------------------------------------------+
 | Argument                                        | Description                                                           |
 +=================================================+=======================================================================+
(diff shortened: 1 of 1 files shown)


docs/drug_metabolite.rst
```diff
@@ -62,42 +66,8 @@ Currently implemented drug metabolite models are:
...
-            base -> s4
-            s4 -> s5
-            s4 -> s6
-    }
-
 .. note::
-    Peripheral compartments are added using the see :ref:`exhaustive stepwise search algorithm<algorithms_modelsearch>`.
-
-Regarding DVID, DVID=1 is connected to the parent drug while DVID=2 is representing the metabolite.
+    Other types of drug metabolite models are not yet fully supported within the tool.
```
stellabelin, line 70: What is meant by "not yet fully supported"? We only support models with two DVIDs where DVID=1 is connected to parent drug and DVID=2 is connected to metabolite right? Was thinking this phrasing might be a bit confusing.

JohanWBorg: Yes and the phrasing is a bit confusing. The parsing of drug metabolite models created outside of pharmpy should work, but require the drug metabolite compartment to be named METABOLITE for us to find it currently. And even though we can parse models with, for example, a metabolite conversion <100%, these cannot be created within pharmpy, which is what I meant by "not yet fully supported".

But a note along the lines of

`Drug metabolite models created outside of pharmpy currently require the metabolite compartment to be named 'METABOLITE' in order to function properly`

might fit the description better?

stellabelin: Yes, I think that's clearer!

PR #3044, opened 2024-05-22
Store input/final model within each tool

JohanWBorg:
When storing the input and final model within `create_results`, sometimes the wrong model was stored (as some of the tools manipulate the final model, for instance covsearch). Now, each tool stores the input model and final model themselves.

### DIFF
--- a/src/pharmpy/tools/covsearch/tool.py
+++ b/src/pharmpy/tools/covsearch/tool.py
@@ -179,7 +179,9 @@ def create_workflow(
     wb = WorkflowBuilder(name=NAME_WF)
 
     # FIXME : Handle when model is None
-    start_task = Task("create_modelentry", _start, model, results, max_eval)
+    store_task = Task("store_input_model", _store_input_model, model, results, max_eval)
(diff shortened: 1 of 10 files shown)


src/pharmpy/tools/covsearch/tool.py
```diff
@@ -222,7 +224,14 @@ def create_workflow(
     return Workflow(wb)
 
 
+def _store_input_model(context, model, results, max_eval):
+    # Create links to input model
+    context.store_input_model_entry(ModelEntry.create(model=model, modelfit_results=results))
+    return max_eval
```
rikardn, line 230: Why pipe `max_eval` through this function? Couldn't it go directly to `_start`?

rikardn (approved):
One comment, but merging as-is anyway

src/pharmpy/tools/covsearch/tool.py
JohanWBorg: This was a workaround to avoid having multiple endpoint tasks, so this way, `_store_input_model` became a part of the workflow.

PR #3370, opened 2024-10-14
First Version of SAMBA approach

pengmanhht:


src/pharmpy/tools/covsearch/tool.py
```diff
@@ -266,91 +315,49 @@ def filter_search_space_and_model(search_space, model):
...
-    # Remove all covariates not part of the search space
-    to_be_removed = model_mfl - ss_mfl
-    covariate_list = to_be_removed.mfl_statement_list(["covariate"])
+    # Clean up all covariate effect in model
+    model_mfl = ModelFeatures.create_from_mfl_string(get_model_features(filtered_model))
+    # covariate effects not in search space, should be kept as it is
+    covariate_to_keep = model_mfl - ss_mfl
+    # covariate effects in both model and search space, should be removed for exploration in future searching steps
+    covariate_to_remove = model_mfl - covariate_to_keep
+    covariate_to_remove = covariate_to_remove.mfl_statement_list(["covariate"])
```
rikardn, line 325: Are covariate effects not removed any longer with this? It is intentional to remove covariate effects not in the search space.

test-requirements.txt
```diff
@@ -6,4 +6,5 @@ pluggy==1.5.0
 pytest==8.3.3
 pytest-asyncio==0.24.0
 pytest-xdist==3.6.1
+statsmodels==0.14.2
```
rikardn, line 9: This shouldn't be in here since it is already a dependency of Pharmpy itself.

rikardn (changes requested):
Most things are small and I could work on fixing them. I had one question about the changes to the `tool.py`

src/pharmpy/modeling/expressions.py
```diff
@@ -329,21 +332,100 @@ def mu_reference_model(model: Model):
...
+    mu_regex = re.compile(r'^mu_\d*$', re.IGNORECASE)
+    for ind_param in ind_parameters:
+        ind_statement = model.statements.find_assignment(ind_param)
+        if not any(re.match(mu_regex, str(p)) for p in ind_statement.free_symbols):
+            return False
+
+    return True
+
+
+def mu_connected_to_parameter(model: Model, parameter: str):
```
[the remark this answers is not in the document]
rikardn: I renamed it to `get_mu_connected_to_parameter`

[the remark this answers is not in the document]
rikardn: I mean `-> str`

src/pharmpy/tools/covsearch/tool.py
rikardn: After discussion we should keep this. Problems with allometric scaling was solved, but instead now covariate effects not in the search space, but in the model will be kept. We should fix that problem later.

PR #3541, opened 2024-12-20
Updated version of SAMBA

pengmanhht:
Update the linear covariate screening of SAMBA method
- mimic the linear covariate selection of Monolix: implement linear covariate selection as stepwise forward inclusion, and use corrected BIC as default selection criterion
- _samba_max_covariates controls the maximum number of covariates selected for each parameter during stepwise linear covariate screening
- _samba_selection_criterion sets the criterion for linear and nonlinear model selection. BIC and LRT are supported.
- _samba_ilnreg_method sets the method for linear covariate screening. OLS, WLS, and LME are supported.
- add loginfo showing the linear and nonlinear model selection results


src/pharmpy/tools/run.py
```diff
@@ -954,7 +954,7 @@ def __gt__(self, value):
 def is_strictness_fulfilled(
     model: Model,
     results: ModelfitResults,
-    strictness: str,
+    strictness: Optional[str],
```
rikardn, line 957: The `strictness` here is intentionally no longer optional. Each tool should have the default strictness string.

src/pharmpy/tools/covsearch/tool.py
```diff
@@ -127,16 +127,17 @@ def create_workflow(
...
     ] = 'scm-forward-then-backward',
     max_eval: bool = False,
     adaptive_scope_reduction: bool = False,
     strictness: Optional[str] = "minimization_successful or (rounding_errors and sigdigs>=0.1)",
     naming_index_offset: Optional[int] = 0,
-    nsamples: int = 10,
-    _statsmodels: bool = False,
-    _weighted_linreg: bool = False,
-    _lin_filter: int = 0,
+    samba_nsamples: int = 10,
```
rikardn, line 136: I think keeping the `nsamples` name has the benefit of being more general in case there are other non-samba sample based methods.

rikardn (changes requested):
Thanks for the PR! I have some comments

src/pharmpy/tools/covsearch/tool.py
pengmanhht: Thank you. I'll rename the option.

src/pharmpy/tools/run.py
pengmanhht: Thank you for the clarification. I'll modify the `strictness` in this run.py and in covsearch tool.

src/pharmpy/tools/covsearch/tool.py
rikardn: Great. You can check how it is done in for example `iivsearch`

rikardn (approved):
Lookgs good to me now. Thanks for the updates!

PR #3640, opened 2025-02-13
fix: Add Windows support for Rscript executable in get_rpath function

ezraluuu:
This pull request addresses an issue where the script was failing to locate the Rscript executable on Windows. On Windows, the executable is named Rscript.exe, but the script was checking for Rscript, which caused path.is_file() to return False.


rikardn:
Thanks for the PR! I fixed the small linting error.

PR #3732, opened 2025-04-09
fix typo amd.rst

leovuong:


rikardn:
Thanks!

leovuong:
it's weird that I'm not on the list of contributors 😄

rikardn:
Yes, I have seen this myself in a case where I was using a different email in the commit from what I use on github, so github couldn't  detect a connection between my commit and my account. I didn't look into how to solve it.

leovuong:
thanks for the explanation. I will pay more attention next time.

PR #3795, opened 2025-05-14
Add linux and windows on arm for unit testing in CI

rikardn:


rikardn:
We now test Linux on ARM. Windows on ARM will need more work. numpy have no wheels and the numpy build fails. Closing this for now.

PR #3819, opened 2025-06-03
fix typos

leovuong:
fix typos on the "Automatic Model Development (AMD)" page


rikardn:
Good catch! Thanks for fixing.

PR #4032, opened 2025-10-06
Avoid parsing the same model multiple times in tests

king-of-poppk:
By using `parse_modelfit_results` instead of `read_modelfit_results`.

This reduces tests running time by ~20% on 8 worker threads.

### DIFF
--- a/tests/modeling/test_common.py
+++ b/tests/modeling/test_common.py
@@ -36,7 +36,7 @@ from pharmpy.modeling import (
     split_joint_distribution,
     write_model,
 )
-from pharmpy.tools import read_modelfit_results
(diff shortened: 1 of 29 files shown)


king-of-poppk:
@rikardn There is one unit test left here: https://github.com/pharmpy/pharmpy/blob/50d3be80fb7b6265bcc71c4981710bba986fb2b4/tests/nonmem/test_modelfit_results.py#L285

There is no test for `~/` resolution (there is one for `read_model`).

There are also some integration tests that use `read_modelfit_results`.

Some tools use `read_modelfit_results`, but I am coming with another PR for that.

PR #4034, opened 2025-10-06
Split `read_nonmem_dataset` tests on small inputs

king-of-poppk:


rikardn (approved):
This looks good. Makes it clearer what we are testing for and makes parallelization possible.

PR #4037, opened 2025-10-06
Add dev container configuration for VS Code

king-of-poppk:
This is a first draft that works for me. Includes Python 3.11 and 3.12.


rikardn (approved):
I haven't reviewed this in detail

PR #4038, opened 2025-10-06
Include tests requirements in type-checking `tox` environment

king-of-poppk:
This improves type checking experience in VS Code.


king-of-poppk:
This change would also be required anyway if test files were checked for type issues.

rikardn:
Ok. Perhaps add a comment so that we don't remove that line thinking it is unnecessary.

PR #4039, opened 2025-10-06
Improve `NONMEMTableFile` performance

king-of-poppk:
We only replace `*OBJ` in first header line.

### DIFF
--- a/src/pharmpy/model/external/nonmem/table.py
+++ b/src/pharmpy/model/external/nonmem/table.py
@@ -9,6 +9,8 @@ from pharmpy.deps import numpy as np
 from pharmpy.deps import pandas as pd
 from pharmpy.internals.math import flattened_to_symmetric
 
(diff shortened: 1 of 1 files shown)


src/pharmpy/model/external/nonmem/table.py
```diff
@@ -61,8 +63,8 @@ def _parse_table(
 
         table_line = None if notitle else content.pop(0)
 
+        content[0] = OBJ.sub("OBJ", content[0])
```
king-of-poppk, line 66: Original offense commited in https://github.com/pharmpy/pharmpy/commit/e05ace3e4ab2ceac427a0821586ea14067d937a2 when an attempt was made at generalizing code that only operated on column names to operate on entire table contents instead due to the lack of a trivial way to do column name substitution based on a regular expression.

Maybe a better solution exists?

rikardn (approved):
Good catch!

PR #4040, opened 2025-10-06
Speed up large dataset parsing with `engine='c'`

king-of-poppk:


rikardn (approved):
Thank you. I think this code is very clear. How much faster was this version?

king-of-poppk:
> Thank you. I think this code is very clear. How much faster was this version?

**BEFORE**

```py
from pathlib import Path

from pharmpy.model.external.nonmem.nmtran_parser import NMTranParser
from pharmpy.model.external.nonmem.parsing import parse_datainfo, parse_dataset

def elapsed(fn, *args, **kwargs):
  from time import time
  from logging import getLogger
  logger = getLogger(__name__)
  begin = time()
  x = fn(*args, **kwargs)
  end = time()
  logger.error(end-begin)
  return x

path = "..."  # NOTE: 52MB, ~430k rows, ~65k rows when filtered

with open(path, encoding = "latin-1") as fp:
  code = fp.read()

cs = NMTranParser().parse(code)
di = parse_datainfo(cs, Path(path))

elapsed(parse_dataset, di, cs, raw=False)   # NOTE: Best of 3 after warmup: ~6.3s
```

**AFTER**

```py
elapsed(parse_dataset, di, cs, raw=False)   # NOTE: Best of 3 after warmup: ~3.5s
```

PR #4046, opened 2025-10-08
Speed up large dataset parsing by performing filtering and conversion in two passes

king-of-poppk:
We perform filtering and conversion in at most two passes.

Conversion passes are interleaved with filtering passes to avoid converting entries that get excluded anyway.

Conversion is slightly optimized by performing multiple-column conversions with a single `apply` call by vectorizing `_convert_data_item`.

We also allow the user to control which columns should be converted in the result dataframe when `raw=False`. We ensure that columns necessary for numeric filtering are converted before that filtering pass, and "unconvert" those afterwards if the user did not request for them to be converted in result.

### DIFF
--- a/src/pharmpy/model/external/nonmem/dataset.py
+++ b/src/pharmpy/model/external/nonmem/dataset.py
@@ -10,9 +10,9 @@ from pharmpy.model import DatasetError, DatasetWarning
 
 from .nmtran_data import SEP_INPUT, NMTRANDataIO, read_NMTRAN_data
 from .nmtran_filter import (
-    Filter,
     filter_column,
     filter_dataset,
+    filter_update_in_place,
     operator_type,
     parse_filter_statements,
 )
@@ -95,18 +95,6 @@ def _idcol(df: pd.DataFrame):
     return next(filter(df.columns.__contains__, ("ID", "L1")), None)
 
 
-def _filters_original(filters: Iterable[Filter]):
-    for f in filters:
-        if operator_type(f.operator) is str:
-            yield f
-
-
-def _filters_parsed(filters: Iterable[Filter]):
-    for f in filters:
-        if operator_type(f.operator) is not str:
-            yield f
-
-
 def read_nonmem_dataset(
     path_or_io,
     raw=False,
@@ -186,21 +174,19 @@ def read_nonmem_dataset(
     original = df
     parsed = original.copy()
     _parsed_for_filters = set()
-    if filters:
+
+    for filter in filters:
         negate = statements is ignore
+        df = original if operator_type(filter.operator) is str else parsed
 
-        filters_original = list(_filters_original(filters))
-        if filters_original:
-            filter_dataset(original, parsed, filters_original, negate)
+        if df is parsed:
+            if (column := filter_column(filter)) not in _parsed_for_filters:
+                parsed[[column]] = convert(parsed[[column]], str(null_value), missing_data_token)
+                _parsed_for_filters.add(column)
 
-        filters_parsed = list(_filters_parsed(filters))
-        if filters_parsed:
-            _parsed_for_filters = set(map(filter_column, filters_parsed))
-            if _parsed_for_filters:
-                _cols = list(_parsed_for_filters)
-                parsed[_cols] = convert(parsed[_cols], str(null_value), missing_data_token)
-
-            filter_dataset(original, parsed, filters_parsed, negate)
+        res = filter_dataset(df, [filter], negate)
+        filter_update_in_place(original, res)
+        filter_update_in_place(parsed, res)
 
     _parse_columns = (
         set(parse_columns)
--- a/src/pharmpy/model/external/nonmem/nmtran_filter.py
+++ b/src/pharmpy/model/external/nonmem/nmtran_filter.py
@@ -1,5 +1,4 @@
 from dataclasses import dataclass
-from itertools import chain
 from typing import Iterable, Literal, Union, cast
 
 from lark import Lark, Token
@@ -124,9 +123,9 @@ def _filter_to_expr(filter: Filter):
     _operator = operator_symbol(filter.operator)
 
     if _operator_type is str:
-        return f'(o{filter.column} {_operator} "{filter.expr}")'
+        return f'({filter.column} {_operator} "{filter.expr}")'
     else:
-        return f"(p{filter.column} {_operator} {filter.expr})"
+        return f"({filter.column} {_operator} {filter.expr})"
 
 
 def _negation(expression: str):
@@ -137,29 +136,13 @@ def _conjunction(expressions: Iterable[str]):
     return " & ".join(expressions)
 
 
-def _query(df: pd.DataFrame, resolver: dict[str, pd.Series], expr: str):
-    from pandas.core.computation.eval import eval as _eval
-
-    return cast(
-        pd.Series,
-        _eval(
-            expr,
-            level=0,
-            local_dict={},
-            global_dict={},
-            resolvers=[resolver],
-        ),
-    )
-
-
-def _filter_update_in_place(df: pd.DataFrame, res: pd.Series):
+def filter_update_in_place(df: pd.DataFrame, res: pd.Series):
     view = df.loc[res]
     df._update_inplace(view)
 
 
 def filter_dataset(
-    original: pd.DataFrame,
-    parsed: pd.DataFrame,
+    df: pd.DataFrame,
     filters: Iterable[Filter],
     negate: bool,
 ):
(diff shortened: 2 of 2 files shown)


tests/nonmem/test_read.py
```diff
@@ -169,10 +169,13 @@ def test_nonmem_dataset_with_ignore_accept_case_2():
...
-        read_nonmem_dataset(
-            StringIO("1,2\n1,3\n2,4\n2,a"), colnames=colnames, ignore=['DV.EQN.2', 'DV.EQ.a']
-        )
+    df = read_nonmem_dataset(
+        StringIO("1,2\n1,3\n2,4\n2,a"), colnames=colnames, ignore=['DV.EQN.2', 'DV.EQ.a']
+    )
+    assert len(df) == 2
+    assert list(df.columns) == colnames
+    assert list(df.iloc[0]) == [1, 3]
+    assert list(df.iloc[1]) == [2, 4]
```
king-of-poppk, line 178: Is it a feature or a bug that we now are able to handle such datasets? Does NONMEM really perform filtering in that exact order? And does it crash? Does it matter if we handle more things than NONMEM?

src/pharmpy/model/external/nonmem/nmtran_filter.py
```diff
@@ -0,0 +1,188 @@
...
+            ((f"o{key}", value) for key, value in original._get_cleaned_column_resolvers().items()),
+            ((f"p{key}", value) for key, value in parsed._get_index_resolvers().items()),
+            ((f"p{key}", value) for key, value in parsed._get_cleaned_column_resolvers().items()),
+        )
+    )
+
+    res = _query(parsed, resolver, expr)
+
+    _filter_update_in_place(original, res)
+    _filter_update_in_place(parsed, res)
```
king-of-poppk, line 188: This can be simplified further if we will never need to apply `str` and non-`str` filters simultaneously. Then filtering can be done on `original` first, then on `parsed`. Originally, we thought it would be possible to do both simultaneously, but this does not work in cases where `str`-filtering excludes values in non-`str`-filtering columns.

I also now realize that since we don't filter one column at a time, the same can happen with non-`str`-filtering only, so maybe it is unreasonable to assume this would always work? What does NONMEM really allow?

One approach we could take, is to first attempt filtering in one pass, then two passes, and finally, one column at a time.

king-of-poppk: It is possible that filtering one column at a time could be more efficient in cases where conversion works on non-filtered data but filtering on the first few column would make the remaining data so small that it speeds up conversion of the remaining columns we need to filter on. So maybe this is not a good idea in general unless it is supplemented by some heuristic and different implementation strategies. For instance:

  - Apply equality filters first since they likely exclude a lot of data
  - Have some thresholds depending on number of rows remaining vs number of filtered columns remaining

Another idea is to use the `pyarrow` parser. We have a working PoC, but the problem is that that parser does not allow keeping the original string data, however, it is tremendously fast at conversion (only works for a subset of the Fortran number syntax). So maybe an idea is to do multiple passes of a "partial" conversion which is works only for columns that respect a certain format, interleaved with filtering passes. When no more remaining filtered-on columns can be converted with this approach, then try the slow conversion.

src/pharmpy/model/external/nonmem/nmtran_filter.py
```diff
@@ -0,0 +1,188 @@
...
+    if negate:
+        expressions = map(_negation, expressions)
+
+    expr = _conjunction(expressions)
+
+    resolver: dict[str, pd.Series] = dict(
+        chain(
+            (
+                (f"o{key}", value)
+                for key, value in original._get_index_resolvers().items()  # type: ignore
```
rikardn: Is this calling a private method on the pandas `DataFrame`? If so it feels like something we shouldn't do.

src/pharmpy/model/external/nonmem/dataset.py
```diff
@@ -61,129 +67,42 @@ def _convert_data_item(x, null_value, missing_data_token):
...
+            DatasetWarning,
+        )
+        df[idcol] = id_change.cumsum()
 
 
-def _filter_ignore_accept(df, ignore, accept, null_value, missing_data_token):
-    if ignore and accept:
-        raise ValueError("Cannot have both IGNORE and ACCEPT")
+def _idcol(df: pd.DataFrame):
+    return next(filter(df.columns.__contains__, ("ID", "L1")), None)
```
rikardn: `__contains__` here is technically private. The use of `None` here is quite obscure. I would keep the original code which to me is much clearer.

```python
if 'ID' in df.columns:
    label = 'ID'
elif 'L1' in df.columns:
    label = 'L1'
else:
    label = None
```

king-of-poppk: OK. Should not make a lot of difference with respect to speed. If yes, `if`s should probably be faster in this particular case.

I don't think however that "official" "dunder" methods are considered private (although they technically start with `_` like private methods and properties) since they are part of the language "contract" (`with`, `next`, `iter`, ...). Pyright does not consider those methods private.

src/pharmpy/model/external/nonmem/nmtran_filter.py
rikardn: It is a maintenance risk to use private methods. Please replace.

src/pharmpy/model/external/nonmem/dataset.py
```diff
@@ -251,47 +170,82 @@ def read_nonmem_dataset(
...
+
+    statements = ignore or accept
+    if statements is None:
+        filters = []
+    else:
+        filters = list(parse_filter_statements(statements))
+
+    original = df
+    parsed = original.copy()
+    _parsed_for_filters = set()
```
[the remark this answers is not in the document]
king-of-poppk: I will see whether those are necessary in a final refactoring pass.

src/pharmpy/model/external/nonmem/nmtran_filter.py
rikardn: Ok, so after this PR you think the parsing is fast enough for your use cases? In that case we don't need to complicate things. If we want to go even faster we have more ideas (like multiple selectable parsers, option to only parse bare minimum from dataset, and perhaps others), but let us discuss them only if you find it worth while for your use case.

king-of-poppk: To be precise, with this PR, the "conversion" part (string to numeric) is not what takes time anymore in my use cases, because I can select exactly what needs to be converted.

The "filtering" part is also negligible with this PR because we avoid creating a lot of temporary dataframes and series.

After this PR gets merged, the main bottleneck on "large" data is the line separator handling (a single line of code):

https://github.com/pharmpy/pharmpy/blob/65b94595bf6f8e9c86abb2c93c0c25f71edf9989/src/pharmpy/model/external/nonmem/nmtran_data.py#L49

The last changes I need that I have not yet pushed to Pharmpy is the ability to derive the raw dataset and the filtered dataset from the same dataframe, so I only need to call the CParser once per model. Not even sure any changes are needed anymore, need to check. But that's for another PR/discussion.

king-of-poppk: > I also now realize that since we don't filter one column at a time, the same can happen with non-`str`-filtering only, so maybe it is unreasonable to assume this would always work? What does NONMEM really allow?

@rikardn I am converting this PR to a draft for now. I will first add tests that should pass on `main` but not on this PR. Then fix this PR.

king-of-poppk: For reference, the failing tests are:

```console
FAILED tests/nonmem/test_read.py::test_nonmem_dataset_with_ignore_accept_case_3 - Failed: DID NOT RAISE <class 'pharmpy.model.data.DatasetError'>
FAILED tests/nonmem/test_read.py::test_nonmem_dataset_with_numeric_ignore_filter_excluding_character_data - pharmpy.model.data.DatasetError: Could not convert the fortran number a to float
FAILED tests/nonmem/test_read.py::test_nonmem_dataset_with_numeric_accept_filter_excluding_character_data - pharmpy.model.data.DatasetError: Could not convert the fortran number a to float
FAILED tests/nonmem/test_read.py::test_nonmem_dataset_with_interleaved_numeric_ignore_filter_and_final_string_filter - Failed: DID NOT RAISE <class 'pharmpy.model.data.DatasetError'>
FAILED tests/nonmem/test_read.py::test_nonmem_dataset_with_interleaved_numeric_accept_filter_and_final_string_filter - Failed: DID NOT RAISE <class 'pharmpy.model.data.DatasetError'>
```

king-of-poppk:
I have restored the original behavior. Now all tests are passing.

> I've extended testing to additional data I am interested in (including the 52MB dataset referenced in #4040). My gains there average at ~18% on medium to large data (~40k - ~430k rows), and ~10% on small data (~1k rows)

Performance for small data has not changed. For medium and large data the gain went down to ~15%.

king-of-poppk:
I've made some significant progress on other fronts too so I suggest we first look at those more clear-cut "improvements" and then go back to the ideas here in a new PR with a different performance reference.

I have also worked on a profiling solution for Pharmpy that should help better identify these performance improvements.

But, without spoiling it, let's say I have a solution for data filtering and conversion that adresses all our concerns (100% NONMEM-compliant, according to your analysis of NONMEM 7.4.3 behavior) and that is even faster than the original proposition.

I will close this PR when the new one is up.

king-of-poppk:
- Superseded by #4083. Closing.

src/pharmpy/model/external/nonmem/nmtran_filter.py
king-of-poppk: - Fixed in #4083.

src/pharmpy/model/external/nonmem/dataset.py
king-of-poppk: - Fixed in #4083.

PR #4085, opened 2025-10-22
Optimize dataset column conversion primitives

king-of-poppk:
- We use a small LRU cache which helps speed up small integer conversion
  - We precompute the result of converting `null_value`.
  - We precompile regular expressions to make sure they are always "cached"
  - We replace disjunctions of equality conditions with tuple containment


king-of-poppk:
@rikardn It looks like `docs-build` failed spuriously.

rikardn:
`docs-build` checks that external links are alive. One of them was down for a moment today.

PR #4088, opened 2025-10-22
Remove special windows eventloop setting

rikardn:
Testing if we can do without this now.


rikardn:
This failed. Opened #4089 to track

PR #4092, opened 2025-10-22
Place all `pharmpy.model.external.nonmem.dataset`-related files under the same directory

king-of-poppk:
This is only rewriting/refactoring and does not include any significant performance improvements or regressions.


src/pharmpy/model/external/nonmem/dataset/_columns.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def idcol(df: pd.DataFrame) -> str | None:
+    columns = df.columns
+    if 'ID' in columns:
+        return 'ID'
+    elif 'L1' in columns:
+        return 'L1'
+    else:
+        return None
```
rikardn: A little to little to keep in separate file? Since it is related to how nmtran sees the dataset perhaps it could be placed there.

rikardn: I think there is other machinery that parses `$INPUT` to get the "true" names of the columns. It might be possible to write `$INPUT ID=MYID`, but that should be handled (not entirely sure without testing though)

rikardn (approved):

PR #4098, opened 2025-10-23
Only convert ID column once in `read_nonmem_dataset`

king-of-poppk:


rikardn (approved):
This was surprisingly low hanging fruit. And the walrus got to join the party (turns out that there is no walrus emoji). I guess assigning `_ids` before the if statement could be an option.

So: LGTM

king-of-poppk:
> I guess assigning `_ids` before the if statement could be an option.

The original refactoring was just that. Then I realized we do not need `_ids` when `raw=True`.

PR #4099, opened 2025-10-23
Publicize low-level API for `parse_model` and `read_nonmem_dataset`

king-of-poppk:
- To be rebased on #4097 before merging.

PR #4105, opened 2025-10-27
Add script to benchmark and profile NONMEM's `parse_model`

king-of-poppk:
- ~To be rebased on #4099.~ **EDIT**: DONE.

To run it without profiling:

```sh
tox -e plain -- python scripts/benchmark/model/external/nonmem/parse_model.py tests/testdata/ src/pharmpy/internals/example_models/
```

To run it with profiling output in `prof/`:

```sh
tox -e instrument -- python scripts/benchmark/model/external/nonmem/parse_model.py tests/testdata/ src/pharmpy/internals/example_models/ -p prof/
```

### DIFF
--- a/scripts/benchmark/model/external/nonmem/parse_model.py
+++ b/scripts/benchmark/model/external/nonmem/parse_model.py
@@ -1,0 +1,239 @@
+import re
+import sys
+
+from contextlib import contextmanager
+from dataclasses import dataclass
+from datetime import datetime, timedelta
+from pathlib import Path, PurePath
+from typing import Iterable, Literal, Optional, Union, cast
+
+from pharmpy.deps import pandas as pd
+
+from pharmpy.model.datainfo import DataInfo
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -86,6 +86,16 @@ deps =
     -rrequirements.txt
 commands = {posargs}
 
+
+[testenv:{py311-,py312-,py313-,py314-,}instrument]
```
rikardn: We have a `profile` env using the pytest profiling. Couldn't that be used? Is this different or better?

king-of-poppk: The `profile` env uses a dependency to profile tests from `pytest`. It is not suited for anything else. Maybe it should always be referred to as `unit-profile`. Not sure why `unit` is optional.

rikardn: Ok, thanks!

rikardn (approved):

PR #4106, opened 2025-10-27
Fix bug when dataset includes individuals without observations

king-of-poppk:
@rikardn How can we cover this for regressions?


rikardn (commented):
Was this bug always triggered when we had removed ID with no observation?

king-of-poppk:
> That would be great. You could base it on pheno.

This one? https://github.com/pharmpy/pharmpy/blob/main/src/pharmpy/internals/example_models/pheno.mod

PR #4107, opened 2025-10-27
Fix implementation of `subs` to handle `Atom`s with `args`

king-of-poppk:
An example of an `Atom` with `args` is `Quantity` which we use for units.

### DIFF
--- a/src/pharmpy/internals/expr/subs.py
+++ b/src/pharmpy/internals/expr/subs.py
@@ -56,7 +56,8 @@ def _subs_atoms_simultaneously(
     subs_new_args: Callable[[sympy.Expr, list[sympy.Expr]], sympy.Expr], expr: sympy.Expr
 ):
     stack = [expr]
-    output = [[], []]
+    # NOTE: Bypass substitution of atom arguments
+    output = [[], expr.args if isinstance(expr, sympy.Atom) else []]
 
     while stack:
         e = stack[-1]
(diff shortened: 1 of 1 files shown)


rikardn:
For expressions we mostly rely on `Expr.subs`. There are some remnants of using the old `subs` function. Is this only needed for units substitutions?

king-of-poppk:
Since `Unit` is currently based on `sympy`:

https://github.com/pharmpy/pharmpy/blob/7bd49c039c1e498759f50ec0a7cf01125fb8baf3/src/pharmpy/basic/unit.py#L15

We created this PR to enable the following dependent optimization: https://github.com/pharmpy/pharmpy/pull/4112/files

Of course we could compare that to using `symengine` for `Unit`, if there is no constraint of using `sympy` there.


This PR also prepares `subs` for future uses (when using `Atom`s that have non-empty `args`). For reference, `subs` being fixed here is still currently used in:

  - https://github.com/pharmpy/pharmpy/blob/63b20b1fb4df275ed13132b819801fd125f5e5a4/src/pharmpy/internals/math.py#L103
  - https://github.com/pharmpy/pharmpy/blob/63b20b1fb4df275ed13132b819801fd125f5e5a4/src/pharmpy/model/random_variables.py#L960
  - https://github.com/pharmpy/pharmpy/blob/63b20b1fb4df275ed13132b819801fd125f5e5a4/src/pharmpy/modeling/covariate_effect.py#L474
  - https://github.com/pharmpy/pharmpy/blob/63b20b1fb4df275ed13132b819801fd125f5e5a4/src/pharmpy/modeling/expressions.py#L348
  - https://github.com/pharmpy/pharmpy/blob/63b20b1fb4df275ed13132b819801fd125f5e5a4/src/pharmpy/modeling/expressions.py#L375-L376
  - https://github.com/pharmpy/pharmpy/blob/63b20b1fb4df275ed13132b819801fd125f5e5a4/src/pharmpy/modeling/expressions.py#L491-L496
  - https://github.com/pharmpy/pharmpy/blob/63b20b1fb4df275ed13132b819801fd125f5e5a4/src/pharmpy/modeling/results.py#L380-L390
  - https://github.com/pharmpy/pharmpy/blob/63b20b1fb4df275ed13132b819801fd125f5e5a4/src/pharmpy/modeling/results.py#L487
  - https://github.com/pharmpy/pharmpy/blob/63b20b1fb4df275ed13132b819801fd125f5e5a4/src/pharmpy/modeling/units.py#L78-L85

PR #4108, opened 2025-10-27
Skip scanning for warnings and errors in `NONMEMResultsFile` when `log=None`

king-of-poppk:
This does not require profiling or benchmarking in my opinion since the benefit is obvious.


rikardn (approved):
LGTM

PR #4109, opened 2025-10-27
Speed up lookup of last assignment for a given `Symbol` in `Statements`

king-of-poppk:
This does not require profiling or benchmarking in my opinion since the benefit is obvious.


rikardn (approved):
Both clearer and faster so LGTM

PR #4121, opened 2025-10-29
Show both Wall and CPU time in benchmark script for `parse_model`

king-of-poppk:


king-of-poppk:
@rikardn This suffered a spurious failure on GHA it seems.

PR #4150, opened 2025-11-12
Optimize Hermitian matrix PSD checks

king-of-poppk:
@rikardn Waiting on your feedback for the approach in general.

### DIFF
--- a/src/pharmpy/internals/math.py
+++ b/src/pharmpy/internals/math.py
@@ -1,4 +1,6 @@
 import math
+from itertools import count
+from typing import Optional
 
 from pharmpy.deps import numpy as np
 from pharmpy.deps import sympy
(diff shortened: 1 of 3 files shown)


src/pharmpy/internals/math.py
```diff
@@ -105,10 +106,18 @@ def se_delta_method(expr, values, cov):
     return se
 
 
-def is_positive_semidefinite(A):
+def is_symmetric(A):
+    return (A == A.T).all().all()
+
+
+def is_positive_semidefinite(A, eigvals=None):
```
rikardn: Having the function as argument here creates a coupling. Perhaps a different function for known symmetric matrices?

rikardn (approved):
LGTM!

Issue #1872, opened 2023-07-19
AssertionError when loading example model


keithnewman:
### Bug report

When attempting to load the "pheno" or "pheno-linear" example models in pharmpy, using the [example in the documentation](https://github.com/pharmpy/pharmpy/blob/3033e652b74f69cb0e9dcf9f026d4c3b6951bce2/src/pharmpy/modeling/common.py#L344-L345) for the method, an assertion error occurs. The same error message appears when attempting to load some of my own model files.

```
$ pip3 install pharmpy-core
$ python3
Python 3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pharmpy
>>> pharmpy.__version__
'0.97.0'
>>> from pharmpy.modeling import load_example_model
>>> mod=load_example_model("pheno")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/[censored_path]/lib/python3.10/site-packages/pharmpy/modeling/common.py", line 381, in load_example_model
    model = read_model(path)
  File "/[censored_path]/lib/python3.10/site-packages/pharmpy/modeling/common.py", line 51, in read_model
    model = Model.parse_model(path)
  File "/[censored_path]/lib/python3.10/site-packages/pharmpy/model/model.py", line 640, in parse_model
    model = model_module.parse_model(code, path)
  File "/[censored_path]/lib/python3.10/site-packages/pharmpy/model/external/nonmem/model.py", line 313, in parse_model
    control_stream = parser.parse(code)
  File "/[censored_path]/lib/python3.10/site-packages/pharmpy/model/external/nonmem/nmtran_parser.py", line 46, in parse
    record = create_record(separator + s)
  File "/[censored_path]/lib/python3.10/site-packages/pharmpy/model/external/nonmem/records/factory.py", line 93, in create_record
    root = record_parser_class(content).root
  File "/[censored_path]/lib/python3.10/site-packages/pharmpy/internals/parse/generic.py", line 383, in __init__
    self.root = self.parse(buf)
  File "/[censored_path]/lib/python3.10/site-packages/pharmpy/internals/parse/generic.py", line 403, in parse
    root = processor(self.buffer, root)
  File "/[censored_path]/lib/python3.10/site-packages/pharmpy/internals/parse/ignored.py", line 109, in with_ignored_tokens
[...]

PR #4167, opened 2025-11-19
Speed up handling of ignored tokens in record parsers

king-of-poppk:
This is simply a matter of reverting the temporary fix for lark 1.1.6 introduced in the following commits:
  - f8c27609a Temporary fix for lark 1.1.6 (see #1872)
  - fedc5abce Temporary fix for lark 1.1.6 (ctd.) (see #1872)
    - Comment: https://github.com/pharmpy/pharmpy/commit/fedc5abce6d94628260bf58b2a498a408b3a6423#commitcomment-170823891

The issue was addressed in lark 1.1.7:
  - https://github.com/lark-parser/lark/issues/1304#issuecomment-1644469611
  - https://github.com/lark-parser/lark/pull/1305
    - https://github.com/lark-parser/lark/pull/1305/commits/5961d720fb26a6632868bd1e783be93c0e57ea52

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ setup(
         'pharmacometrics',
     ],
     install_requires=[
-        'lark>=1.1.4',
+        'lark>=1.1.7',
         'sympy>=1.9',
         'symengine>=0.13.0',
         'pandas>=1.4, !=2.1.0',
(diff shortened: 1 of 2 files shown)


king-of-poppk:
@rikardn I think you should witness a massive improvement on running the test suite with this one.

Next time such a fix is needed in the parsing logic I think at least some benchmarking should be done to evaluate the performance degradation.

In this particular case, the added logic was inserted in a hot "loop" (recursion) and it would have been straightforward to cache the value of `version("lark") == ...`

PR #4182, opened 2025-11-27
Cache expensive properties of `RandomVariables` instances

king-of-poppk:
The original motivation comes from `pharmpy.model.external.nonmem.update.create_name_map` being expensive to compute due to the following line:

https://github.com/pharmpy/pharmpy/blob/794e64224728e166e91edd522185a630beab7d0e/src/pharmpy/model/external/nonmem/update.py#L2421

Initially, I thought I'd just locally cache the computed property, then I realized this construct was used in many places, so it seemed more natural to just make all of those efficient without manual caching:

https://github.com/pharmpy/pharmpy/blob/794e64224728e166e91edd522185a630beab7d0e/src/pharmpy/modeling/estimation.py#L63-L65

https://github.com/pharmpy/pharmpy/blob/794e64224728e166e91edd522185a630beab7d0e/src/pharmpy/modeling/estimation.py#L159-L161

https://github.com/pharmpy/pharmpy/blob/794e64224728e166e91edd522185a630beab7d0e/src/pharmpy/modeling/estimation.py#L171-L173

https://github.com/pharmpy/pharmpy/blob/794e64224728e166e91edd522185a630beab7d0e/src/pharmpy/model/external/nlmixr/ini.py#L19

https://github.com/pharmpy/pharmpy/blob/794e64224728e166e91edd522185a630beab7d0e/src/pharmpy/model/external/nlmixr/error_model.py#L191-L192

https://github.com/pharmpy/pharmpy/blob/794e64224728e166e91edd522185a630beab7d0e/src/pharmpy/model/external/nonmem/update.py#L90

https://github.com/pharmpy/pharmpy/blob/794e64224728e166e91edd522185a630beab7d0e/src/pharmpy/model/external/nonmem/update.py#L2223

I then extended automated caching to properties that look expensive to compute.

PR #4183, opened 2025-11-27
Cache expensive properties of `Distribution` instances

king-of-poppk:
- Extension of #4182 to `Distribution` instances.

### DIFF
--- a/src/pharmpy/model/distributions/symbolic.py
+++ b/src/pharmpy/model/distributions/symbolic.py
@@ -2,12 +2,13 @@ from __future__ import annotations
 
 from abc import abstractmethod
 from collections.abc import Collection, Hashable, Mapping, Sequence, Sized
+from itertools import chain
 from math import sqrt
 from typing import Any, Union
 
 import pharmpy.internals.unicode as unicode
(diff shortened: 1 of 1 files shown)


src/pharmpy/model/distributions/symbolic.py
```diff
@@ -160,6 +160,7 @@ def parameter_names(self) -> tuple[str, ...]:
         return (self._variance.name,)
 
     @property
+    @cache_method_no_args
     def free_symbols(self) -> set[Expr]:
```
king-of-poppk: @rikardn Perhaps this should be a `frozenset` in the future?

rikardn: So we cannot cache mutable properties. In this case I think `frozenset` shouldn't give much problems.

king-of-poppk:
The overhead of the cache is extremely slim. But this PR is just a(n over?)generalization of #4182 which is motivated by the following profile (FREM model):


Screenshot 2025-11-28 at 15 38 17

For just that particular profile, caching `RandomVariables#free_symbols` suffices. My idea was to be proactive here.

PR #4186, opened 2025-11-27
Avoid holding the `.lst` file in memory when parsing it

king-of-poppk:
We still hold a full copy in memory when `log is not None` or when the NONMEM version string cannot be found however. This is addressed in subsequent PRs.

### DIFF
--- a/src/pharmpy/tools/external/nonmem/results_file.py
+++ b/src/pharmpy/tools/external/nonmem/results_file.py
@@ -1,6 +1,8 @@
 import re
+import sys
 from datetime import datetime
-from typing import Iterable, Optional, Union
+from itertools import tee
+from typing import Callable, Generator, Iterable, Iterator, Optional, TypeVar, Union
 
 import dateutil.parser
 from packaging import version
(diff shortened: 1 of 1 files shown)


king-of-poppk:
The tests probably fail on Windows because of `\r`.

src/pharmpy/tools/external/nonmem/results_file.py
```diff
@@ -9,6 +10,41 @@
...
+    try:
+        row = line.decode('utf-8')
+    except UnicodeDecodeError:
+        row = line.decode('latin-1', errors='ignore')
+
+    pos = len(row) - 2
+    if pos >= 0 and row[pos] == '\r':
+        return row[:pos] + '\n'
+    else:
+        return row
```
king-of-poppk: Almost half of the time is spent checking for `\r`, even if there are none. Maybe there is a better way?

```suggestion
    try:
        return line.decode('utf-8')
    except UnicodeDecodeError:
        return line.decode('latin-1', errors='ignore')
```

**BEFORE**

Screenshot 2025-12-02 at 17 01 56


**AFTER**

Screenshot 2025-12-02 at 17 08 58

king-of-poppk:
Some timings on a `.lst` file with ~2M lines:

`main` (ed19d8f5ff57f3d211894bdd909d7b38b5d04b4f)

Screenshot 2025-12-02 at 17 02 25

`king-of-poppk:perf-large-lst-1` (e714daa8a784fa4e396a4bfdc19d7b6ef8842ef6)

Screenshot 2025-12-02 at 17 01 56

src/pharmpy/tools/external/nonmem/results_file.py
```diff
@@ -395,98 +414,81 @@ def log_items(self, lines):
...
-        for row in lines:
-            m = nmversion.match(row)
-            if m:
-                version_number = NONMEMResultsFile.cleanup_version(m.group(1))
-                yield ('nonmem_version', version_number)
-                break  # We will stay at current file position
 
-        self.log_items(lines)
+        with open(path, 'rb') as fp:
+            lines = map(_decode_lst, fp)
```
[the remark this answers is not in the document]
king-of-poppk: > No it is not needed for all lines. We could attempt dividing the lst-file into its parts so that they could be parsed separately. Not all errors/warnings can show up in all parts.

@rikardn I have a subsequent PR that simplifies splitting the LST file into parts and so the approach you are suggesting should work.

src/pharmpy/tools/external/nonmem/results_file.py
king-of-poppk: This will not be a problem anymore if we implement https://github.com/pharmpy/pharmpy/pull/4186#discussion_r2583986041 because the bottleneck is detecting specific line start sequences which are encoded similarly in `utf8` and `latin1`: `b" #"` and (maybe) `b" *"`.

king-of-poppk:
@rikardn This is ready for a final review.

src/pharmpy/tools/external/nonmem/results_file.py
```diff
@@ -9,6 +11,67 @@
...
+    pos = len(row) - 2
+    if pos >= 0 and row[pos] == '\r':
+        return row[:pos] + '\n'
+    else:
+        return row
+
+
+T = TypeVar('T')
+
+_version = tuple(map(int, python_version_tuple()))
```
rikardn: Use sys.platform_info instead. Can be used directly to compare with tuples of ints

rikardn (changes requested):

src/pharmpy/tools/external/nonmem/results_file.py
```diff
@@ -101,6 +164,34 @@ def cleanup_version(v):
             v = '6.0'
         return v
 
+    @staticmethod
+    def read_tere(
+        rows: Iterator[str], lookahead: Callable[[int], Generator[str | None, None, None]]
+    ):
+        read = []
+        while True:
+            preread = next(lookahead(1))
```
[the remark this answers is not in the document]
rikardn: Ok, fair enough.

PR #4205, opened 2025-12-04
Speed up `ExtTable#_get_parameters`

king-of-poppk:
`del ...` is slow for some reason.

### DIFF
--- a/src/pharmpy/model/external/nonmem/table.py
+++ b/src/pharmpy/model/external/nonmem/table.py
@@ -374,10 +374,15 @@ class ExtTable(NONMEMTable):
         row = df.loc[df['ITERATION'] == iteration]
         if row.empty:
             raise KeyError(f'Row {iteration} not available in ext-file')
-        del row['ITERATION']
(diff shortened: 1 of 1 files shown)


src/pharmpy/model/external/nonmem/table.py
```diff
@@ -374,10 +374,15 @@ def _get_parameters(self, iteration, include_thetas=True) -> pd.Series:
...
+
+        keep_cols = set(row.columns)
+        keep_cols.difference_update(('ITERATION', 'OBJ'))
+
         if not include_thetas:
-            row = row[row.columns.drop(list(row.filter(regex='THETA')))]
+            theta_cols = row.filter(regex='THETA', axis=1).columns
+            keep_cols.difference_update(theta_cols)
+
+        row = row[[column for column in row.columns if column in keep_cols]]
```
king-of-poppk: @rikardn This needs to be reinvestigated with pandas v3 since now their default strategy is CoW.

PR #4635, opened 2026-06-15
Use original dataset in NONMEM models whenever possible

stellabelin:
This PR adds support for using the original dataset when executing NONMEM (related to #4424). It will add IGNOREs and DROPs whenever needed and whenever the original dataset could be used, otherwise it will wipe the IGNOREs and use the internal Pharmpy dataset. Additionally, this PR adds a default for common options. 

Some notes/questions:
* Name of the function for getting the old behavior (create a new dataset file) is `reset_dataset`, but perhaps this is too vague? Some other options could be `force_new_dataset`, `use_pharmpy_dataset`, or mimic the common option `always_create_new_dataset_file`
* `reset_dataset` will be exported
* Documentation ongoing in separate branch
* In this implementation the original dataset will be copied to each execute directory, if a user is running e.g. IIVSearch with many models in parallel and a big dataset it would take up a lot of space (but this should be an issue in the previous implementation too)


src/pharmpy/model/external/fcon/model.py
```diff
@@ -42,33 +44,150 @@ def _parse_labels_and_formats(code):
...
+        else:
+            if final_row:
+                rows[-1].append(inner_fmt)
+            else:
+                rows.append([inner_fmt])
+                final_row = True
+    return rows
+
+
+def _replace_linebreak(line):
```
rikardn: So what this function does is replacing / at the outermost parens nesting? Perhaps change the name to better reflect this or add a comment.

src/pharmpy/model/external/fcon/model.py
```diff
@@ -42,33 +44,150 @@ def _parse_labels_and_formats(code):
...
-    df = pd.read_fwf(path.parent / 'FDATA', widths=widths, header=None, names=labels)
+    if len(formats) > 1:
+        df = read_multiline_observations(path.parent / 'FDATA', labels, formats)
+    else:
+        formats = formats[0]
+        for fmt in formats:
+            no_of_cols, col_width = get_format_args(fmt)
+            widths += [col_width] * no_of_cols
+        df = pd.read_fwf(path.parent / 'FDATA', widths=widths, header=None, names=labels)
+    df.index = range(1, len(df) + 1)
```
rikardn: How common is a single line format? Do we need this special case?

src/pharmpy/model/external/fcon/model.py
```diff
@@ -42,33 +44,150 @@ def _parse_labels_and_formats(code):
...
+
+        if no_of_rows > 1:
+            rows.extend([[inner_fmt]] * no_of_rows)
+        else:
+            if final_row:
+                rows[-1].append(inner_fmt)
+            else:
+                rows.append([inner_fmt])
+                final_row = True
+    return rows
```
rikardn: I think it would be better to not assume that NONMEM always will have parenthesed formats first always ending with /. The parser would probably be easier to understand if not and it would be easier to fix other future problems with the parser.

src/pharmpy/model/external/fcon/model.py
```diff
@@ -42,33 +44,150 @@ def _parse_labels_and_formats(code):
...
+        for fmt in formats:
+            no_of_cols, col_width = get_format_args(fmt)
+            widths += [col_width] * no_of_cols
+        df = pd.read_fwf(path.parent / 'FDATA', widths=widths, header=None, names=labels)
+    df.index = range(1, len(df) + 1)
+
+    # Copied conversion from pharmpy/model/external/nonmem/dataset/__init__.py
+    idcol = _idcol(labels)
+    if all((_ids := df[idcol].astype('int32')) == df[idcol]):
+        df[idcol] = _ids
```
rikardn: This is hard to follow. Perhaps break it apart and move into a function. Could be put in internals if needed in multiple places

src/pharmpy/model/external/fcon/model.py
```diff
@@ -42,33 +44,150 @@ def _parse_labels_and_formats(code):
...
+
     return df
 
 
+def get_format_args(fmt):
+    m = FORTRAN_FORMAT.match(fmt)
+    if not m:
+        raise ModelSyntaxError(f"Unrecognized Data format in FORM: {fmt}")
```
rikardn, line 134: If this reaches the user perhaps clarify that it is read from the FCON file

src/pharmpy/model/external/fcon/model.py
```diff
@@ -42,33 +44,150 @@ def _parse_labels_and_formats(code):
...
+    no_of_cols, col_width = int(m.group(1)), int(m.group(2))
+    return no_of_cols, col_width
+
+
+def read_multiline_observations(path, labels, formats):
+    observation = []
```
rikardn: Perhaps name this `row` or `current_row` or `currow` or similar.

src/pharmpy/model/external/nonmem/records/data_record.py
```diff
@@ -159,44 +159,47 @@ def ignore(self):
...
+                    bexp = BooleanExpr.gt(lhs, rhs)
+                elif op == 'OP_LT_EQ':
+                    bexp = BooleanExpr.le(lhs, rhs)
+                else:  # op == 'OP_GT_EQ'
+                    bexp = BooleanExpr.ge(lhs, rhs)
+                if child.rule == 'accept':
+                    bexp = ~bexp
+                ignore = Ignore.create(expression=bexp, strings=strings)
+                filters.append(ignore)
+        return filters
```
rikardn, line 202: We already do the parsing in `parse_filter_statements`

src/pharmpy/modeling/data.py
```diff
@@ -2702,3 +2702,28 @@ def _get_all_statement_deps(model: Model, expr: Expr) -> set[Expr]:
             all_deps |= symbol_deps
         all_deps |= {s}
     return all_deps
+
+
+def reset_dataset(model: Model) -> Model:
+    """Resets dataset so that Pharmpy dataset is used in e.g. estimation
```
rikardn: Add to docstring that it removes the path and the provenance, i.e. the connection to the dataset file.

src/pharmpy/workflows/contexts/local_directory.py
```diff
@@ -262,7 +262,13 @@ def retrieve_log(self, level: Literal['all', 'current', 'lower'] = 'all') -> pd.
     def retrieve_common_options(self) -> dict[str, Any]:
         ctx_top = self.get_top_level_context()
         meta = ctx_top.retrieve_metadata()
-        return meta['common_options']
+        if 'common_options' in meta:
+            options = meta['common_options']
+        else:
+            from pharmpy.workflows.args import get_default_common_options
+
+            options = get_default_common_options()
```
rikardn, line 270: Do we either get all given common options or all defaults?

src/pharmpy/model/model.py
```diff
@@ -371,7 +371,10 @@ def replace(self, **kwargs) -> Self:
         if new_dataset and dataset is not None:
             datainfo = update_datainfo(datainfo, dataset)
             # FIXME: second condition occurs when e.g. ignoring rows, different mechanism will handle this later
```
rikardn: Comment rot?

src/pharmpy/model/model.py
```diff
@@ -371,7 +371,10 @@ def replace(self, **kwargs) -> Self:
...
-            if not new_datainfo or datainfo.provenance != self._datainfo.provenance:
+            if not new_datainfo or (
+                datainfo.path == self._datainfo.path
+                and datainfo.provenance != self._datainfo.provenance
+            ):
```
rikardn: document when path is not None, i.e. when model.dataset can be recreated given information in datainfo including Provenance.

src/pharmpy/model/external/nonmem/model.py
```diff
@@ -478,3 +501,31 @@ def handle_case(datainfo, dataset, statements):
...
+        return False
+    if not isinstance(model.datainfo.provenance[0], ReadDataset):
+        return False
+    for op in model.datainfo.provenance:
+        # FIXME: Needs special handling when generating NONMEM code
+        if isinstance(op, Ignore):
+            if not op.expression.is_relational():
+                return False
+        if isinstance(op, (AddRows, AddColumn)):
+            return False
```
rikardn: merge ifs

src/pharmpy/model/external/nonmem/model.py
```diff
@@ -204,16 +219,29 @@ def update_source(self) -> Model:
...
             model = add_dummy_dv(model)
-            cs = update_input(cs, model)
-
-            # Remove IGNORE/ACCEPT. Could do diff between old dataset and find simple
-            # IGNOREs to add i.e. for filter out certain ID.
-            newdata = newdata.remove_ignore().remove_accept()
-            if (
+            keep_dataset = keep_original_dataset(model)
+            cs = update_input(cs, model, keep_dataset)
+            if model.datainfo.path and model.datainfo.path != model.internals.old_datainfo.path:
```
rikardn: Since this logic is central is it possible to make it clearer? Or to make what is done in each case clearer

src/pharmpy/model/external/nonmem/update.py
```diff
@@ -2355,42 +2355,42 @@ def add_dummy_dv(model: Model) -> Model:
     return model
 
 
-def update_input(control_stream, model: Model):
+def update_input(control_stream, model: Model, keep_dataset: bool):
```
rikardn, line 2358: What does keep_dataset mean?

src/pharmpy/tools/external/nonmem/run.py
```diff
@@ -103,6 +108,18 @@ def execute_model(model_entry, context):
         else:
             time.sleep(1)
 
+    if context.retrieve_common_options().get('validate_dataset', False):
+        try:
+            validate_dataset(model, model_path)
+        except (ValueError, AssertionError) as e:
```
rikardn: Should AssertionError really be picked up?

src/pharmpy/tools/external/nonmem/run.py
```diff
@@ -226,3 +243,18 @@ def create_parafile_and_option(context, path: Path, tmp_path: Optional[Path]) ->
         return f"-parafile={path.name}"
     else:
         return ""
+
+
+def validate_dataset(model, path: Path):
+    try:
+        fcon_model = read_model(path / 'FCON')
+    except Exception:
```
rikardn, line 255: Might hide bugs in the parser.

src/pharmpy/tools/external/nonmem/run.py
```diff
@@ -226,3 +243,18 @@ def create_parafile_and_option(context, path: Path, tmp_path: Optional[Path]) ->
...
+        raise ValueError('Could not parse FCON model to compare datasets')
+    dataset = model.dataset.copy()
```
rikardn: Is copy needed? The drop method might create a new DataFrame

src/pharmpy/workflows/model_database/local_directory.py
```diff
@@ -276,25 +276,34 @@ def store_model(self):
...
+            if model.datainfo.path:
+                shutil.copy2(model.datainfo.path, data_path)
+            else:
+                model = write_dataset(model, path=data_path, force=True)
+
             datainfo = model.datainfo.replace(path=data_path)
             model = model.replace(datainfo=datainfo)
-            model = write_dataset(model, path=data_path, force=True)
+
+            # NOTE: Create the index file at .datasets/.hash/<hash>/<dataset_filename>
```
rikardn, line 304: What is the index file?

src/pharmpy/model/external/fcon/model.py
stellabelin: Agreed, more general implementation in 6e9c365648e2502e17f3ba93bfd292d850e989cb

src/pharmpy/model/external/fcon/model.py
stellabelin: Renamed to `_split_formats` in 6e9c365648e2502e17f3ba93bfd292d850e989cb, since newline is now handled separately this function now splits the format string.

src/pharmpy/model/external/fcon/model.py
stellabelin: Any errors will be caught at NONMEM execute and there it will have a less technical error message.

src/pharmpy/model/external/fcon/model.py
stellabelin: No, the same code can be used for both cases! Removed this as a special case for simplicity of code in 825e2e870e91c81e4f2d2555e9f1e1635fcf91ab.

src/pharmpy/model/external/fcon/model.py
stellabelin: Split into separate function `pharmpy.internals.df` called `safe_convert_column_to_int32` in 9d7956ae879c9ef76b9b23dccf445947933aec27.

src/pharmpy/model/external/fcon/model.py
stellabelin: Agreed that it was unclear, renamed in 825e2e870e91c81e4f2d2555e9f1e1635fcf91ab.

src/pharmpy/model/external/nonmem/model.py
stellabelin: Fixed in e1f582a514e7c67896f3cfebef51b1a00159708a.

src/pharmpy/model/external/nonmem/model.py
stellabelin: Yes I agree, new logic in 4503ce1c9dc66f397997ffc89b97c3b266a922f7.

src/pharmpy/tools/external/nonmem/run.py
stellabelin: No, this would pick up more than we would want, I created a new exception class DatasetValidationError in 0eddb1bce16717e2b9c9379394d5f2c75b57b3af.

src/pharmpy/workflows/model_database/local_directory.py
stellabelin: This is a comment regarding how the hashing is handled, so not related to this PR.

src/pharmpy/tools/external/nonmem/run.py
stellabelin: No, it should be a new DataFrame, removed copy and used `inplace=False` option to explicitly state so in e315d4ad49d5a86307731626be01d8dd4429b536

src/pharmpy/tools/external/nonmem/run.py
stellabelin: When it reaches this point in the execute we're mainly interested in whether it failed or not, and the new exception raied will be caught in the execute function to properly log. To make the error clearer we could have the FORTRAN format in the log message to easier debug (see #4641).

src/pharmpy/model/external/nonmem/update.py
stellabelin: Added a description in the docstring in e1255a60bc4d04b167ddaa1c9fc805c760c1592f.

src/pharmpy/modeling/data.py
stellabelin: Updated docstring in e1255a60bc4d04b167ddaa1c9fc805c760c1592f.

src/pharmpy/model/model.py
stellabelin: Removed in e1255a60bc4d04b167ddaa1c9fc805c760c1592f.

src/pharmpy/model/model.py
stellabelin: Documented path in e1255a60bc4d04b167ddaa1c9fc805c760c1592f, also stored condition in variable for clarity in 6f8ce3cd0f3a6e1682b042e087e8bed2cf33aca8.

src/pharmpy/model/external/nonmem/records/data_record.py
stellabelin: The parsing is redone by `parse_filter_statements`, but the AST for the data record is already created before that point. There is some overlap but to me it seems using the structures there would be more complex than using the tree. However, I do see the downside of the duplication of parsing but I think it needs more rethinking outside of this PR (also since `parse_filter_statements` is needed if a dataset should be possible to parse without a NONMEM model).

src/pharmpy/workflows/contexts/local_directory.py
stellabelin: This is for cases where we created a Context standalone (e.g. in testing), in tool runs this will always be a full directory with user set options and the rest default. This is equivalent to how we handle dispatching options.

rikardn (approved):
Thanks! Now LGTM

Issue #4772, opened 2026-08-25
Support projects for tool runs


stellabelin:
Projects will be a top level abstraction that can contain multiple contexts, but is defined by having the model database (and later other dispatching options). This will allow for users to run tools with the same model database which can be useful e.g. when running a bootstrap with some samples and then wanting to run more samples with the same seed, then those model runs can be reused. This will also simplify batch testing of tools. Tasks that need to be done:

- [x] Add class `Project(...)` in `pharmpy.workflows`. Should create directory and model database
- [x] Support `ModelDatabase` as input argument in class `Context`, create context specific model database if not provided
- [x] Add option `'project'` to `pharmpy.tools.run_tool` and use model database as input when creating `Context` object
- [x] Remove `'context'` from dispatching options
- [x] Add function `pharmpy.tools.open_project(name, ref)`
- [ ] Add basic documentation in user guide

First implementation only needs to include model database feature, but should later be extended to handling other dispatching options.

PR #4794, opened 2026-09-01
New Project class 

stellabelin:
Adds a new Project class where contexts created in said project has a common model database (#4772). This allows for faster runs where different tool runs have some models in common.

### DIFF
--- a/src/pharmpy/tools/run.py
+++ b/src/pharmpy/tools/run.py
@@ -56,7 +56,6 @@ def fit(
     model_or_models: Union[Model, list[Model]],
     esttool: Optional[str] = None,
     name: Optional[str] = None,
-    context: Optional[Context] = None,
     ncores: int = 1,
     validate_dataset: bool = False,
 ) -> Union[ModelfitResults, list[ModelfitResults]]:
@@ -70,8 +69,6 @@ def fit(
         Estimation tool to use. None to use default
     name : str
         Name of run
-    context : Context
-        Run in this context
     ncores : int
         Number of cores to use for estimation
     validate_dataset : bool
@@ -98,18 +95,11 @@ def fit(
         (True, model_or_models) if isinstance(model_or_models, Model) else (False, model_or_models)
     )
 
-    if not context:
-        dispatcher = 'local_serial'
-    else:
-        dispatcher = None
-
     modelfit_results = run_tool(
         'modelfit',
         models,
         esttool=esttool,
         name=name,
-        context=context,
-        dispatcher=dispatcher,
         ncores=ncores,
         validate_dataset=validate_dataset,
     )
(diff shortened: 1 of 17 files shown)


src/pharmpy/tools/run.py
```diff
@@ -716,13 +732,20 @@ def _create_new_subcontext_name(context: Context, tool_name: str) -> str:
...
-    ctx = dispatching_options['context']
-    if ctx is None:
-        from pharmpy.workflows import Context
+    from pharmpy.workflows import Context
+
+    name = _get_name(dispatching_options, tool_name)
+    ref = dispatching_options['ref']
+    if (proj := dispatching_options['project']) is not None:
+        if ref:
+            ref = proj.full_ref + ref
```
rikardn: Can we add here in general? Perhaps a method `get_context_ref` is needed or even a `create_context` method on a project similar to `create_subcontext` for a context.

src/pharmpy/workflows/contexts/local_directory.py
```diff
@@ -81,8 +82,12 @@ def _init_top_path(self):
...
             path = parent.parent
 
-    def _init_model_database(self):
-        self._model_database = LocalModelDirectoryDatabase(self._top_path / '.modeldb')
+    def _init_model_database(self, model_database: Optional[ModelDatabase]):
+        if model_database is None:
+            path = self._top_path / '.modeldb'
+            self._model_database = LocalModelDirectoryDatabase(path)
+        else:
+            self._model_database = model_database
```
rikardn, line 90: How does a a resume get this attribute back?

src/pharmpy/workflows/args.py
```diff
@@ -39,6 +40,9 @@ def split_common_options(
     other_options = {}
     for key, value in d.items():
         if key in all_dispatching_options:
+            if key == 'project':
+                if value and not isinstance(value, Project):
```
rikardn: `value is not None` ?

rikardn (commented):
Some comments/questions

src/pharmpy/workflows/args.py
stellabelin: Yes, this should be more specific! Fixed in c89adfb2f7044e0e69bd816246aabdf7e471558b.

src/pharmpy/tools/run.py
stellabelin: Agreed, added a method `Project.get_context_ref` in 9b89697e7231ca6e6934b7aceeaa7ec5e58b2b43.

src/pharmpy/workflows/contexts/local_directory.py
stellabelin: If `project` option is specified, it will use whichever database connected there and input it whenever a context is created. If you would run resume and instead of project specify `name` and `ref` it would crash with an error when trying to retrieve the input model, since it cannot find the model database.

